PR #3, opened 2018-03-19
Exp/ros extract

Fluci:
Important files you might want to check:

- README.md
- extract_camera_streams.m
- config_example.m

I've also adjusted `undistrect.m` by converting it into a function, so no note worthy changes there. All other files are 1:1 as we got them from Dominik.

I've run the entire bag file from Dominik (2018-01-23-10-24-43.bag) in about 20 minutes. 
I haven't tested the other files from Tilo yet.

As far as I'm aware there are no more bugs, since it outputs the same point cloud as the original script. So, if you find something, just ping me.


itko (commented):
I don't have matlab on my machine. Is there anything you wanted me to look at nonetheless? (Other than readme)

Fluci:
Meh, it's less about code quality (which definitely could be a lot better ...) and more about making sure it works for you, when you will need it in the future.

itko:
@luhug check if it works for you and merge the pull request when ready.

luhug (approved):
Works on
macOS 10.11 MATLAB r2017b
Time: 42 min for 4.7GB file

PR #4, opened 2018-03-20
Interfaces/all

Fluci:
Important Changes:

- I cleaned up the interfaces and made sure they compile.
- I re-thought the Trajectory part as discussed at the last meeting: I propose to create descriptors as discussed but the next step is no longer called `Trajectory-building` but `Matching`. Matching connects descriptors as discussed and returns a map how the new clusters should be connected with the old ones. This connection is no longer expressed by `Trajectory` (it was a badly chosen term), but by `ClusterChain`. I think. this expresses the link nature of what we do better. `Trajectory` still exists, but now, it only holds a list of 3d points. This should be more intuitive.
- Centralization of common type definitions in `types.h`.

Please check that you understand at least your interfaces and tell me if you would change something. 

If you're ok with the changes, remove yourself from the reviewers or merge if last.


lib/generic/point_cloud.h
```diff
@@ -0,0 +1,107 @@
...
+
+#include "types.h"
+
+#include <cstddef>
+#include <vector>
+
+namespace MouseTrack {
+
+
+/// A point clouds holds a list of 3D points. Each point can have additional attributes like color intensity.
```
Fluci: typo

lib/generic/point_cloud.cpp
```diff
@@ -0,0 +1,96 @@
...
+Coordinate& PointCloud::Point::z() {
+    return _cloud._zs[_index];
+}
+
+const Coordinate& PointCloud::Point::z() const {
+    return _cloud._zs[_index];
+}
+
+ColorChannel& PointCloud::Point::intensity() {
+    return _cloud._color[_index];
```
Fluci, line 65: test intensity()

lib/generic/point_cloud.h
```diff
@@ -0,0 +1,107 @@
...
+        /// Write access to color intensity.
+        ColorChannel& intensity();
+
+        /// Read access to color intensity.
+        const ColorChannel& intensity() const;
+    };
+
+    /// Exactly the same as `Point` but it only provides read-only access to the data.
+    class ConstantPoint {
+        // Design note: I tend to keep the hierarchy flat,
```
Fluci: double comment, Point should be enough

lib/generic/trajectory.h
```diff
@@ -0,0 +1,31 @@
...
+/// Maintainer: Felice Serena
+///
+///
+
+#pragma once
+
+#include "types.h"
+
+#include <map>
+#include <memory>
```
Fluci: Deprecated include

lib/matching/matching.h
```diff
@@ -0,0 +1,28 @@
...
+namespace MouseTrack {
+
+class Matching {
+public:
+    /// Given a list of cluster chains, we find the best matching to append the given list of descriptors.
+    /// descriptors: a list of descriptors from a new frame
+    /// chains: a list of cluster chains to which we want to appeand one cluster from descriptors
+    /// returns a vector of size descriptors.size(), which holds the index of the matching chain for each descriptor.
+    /// If no match was found, the vector value is set to -1.
+    //virtual std::vector<long> operator()(const std::vector<std::shared_ptr<const ClusterDescriptor>>& descriptors, const std::vector<ClusterChain>& chains) const = 0;
```
Fluci: rubbish

Fluci (commented):
self review

lib/CMakeLists.txt
```diff
@@ -1,24 +1,40 @@
+# add this path to include path of library files
```
Fluci, line 1: Eigen and C++14 break travis

Fluci: fixed with commit 3660f88

luhug (approved):
good

itko (approved):

PR #5, opened 2018-03-25
Add empty classes for descripting and matching

itko:
Creates mock classes so Felice can proceed with the glue code. For now, I am assuming we will proceed with center of gravity for descripting and nearest neighbour for matching.


itko:
@Fluci this doesn't compile yet due to an error in the nearest neighbour. I can't figure out why it errors though! Help!

lib/matching/nearest_neighbour.h
```diff
@@ -0,0 +1,27 @@
...
+	public:
+		/// Given a list of cluster chains, we find the best matching to append the given list of descriptors.
+		/// descriptors: a list of descriptors from a new frame
+		/// chains: a list of cluster chains to which we want to appeand one cluster from descriptors
+		/// returns a vector of size descriptors.size(), which holds the index of the matching chain for each descriptor.
+		/// If no match was found, the vector value is set to -1.
+		virtual std::vector<long> operator()(const std::vector<std::shared_ptr<const ClusterDescriptor>>& descriptors, const std::vector<ClusterChain>& chains);
+};
+
+
```
Fluci: Missing closing bracket for `MouseTrack` namespace: `} // MouseTrack`, this causes the compilation failure

lib/matching/nearest_neighbour.h
```diff
@@ -0,0 +1,27 @@
+/*
+ * nearest_neighbour.h
+ *
+ *  Created on: Mar 25, 2018
+ *      Author: itto
+ */
+
+#ifndef LIB_MATCHING_NEAREST_NEIGHBOUR_H_
```
Fluci: A more modern variant to header guards would be `#pragma once` (see other header files) But I don't really care.

lib/matching/nearest_neighbour.h
```diff
@@ -0,0 +1,27 @@
+/*
```
Fluci: `/** \file` should allow Doxygen to parse the header block (see other header files)

lib/descripting/cog.cpp
```diff
@@ -0,0 +1,17 @@
+/*
```
Fluci: `/** \file` or `/// \file` allow doxygen to parse the comment (see other header files)

lib/descripting/cog.h
```diff
@@ -0,0 +1,25 @@
+/*
```
Fluci: `/** \file` or `/// \file`

lib/descripting/cog.h
```diff
@@ -0,0 +1,25 @@
...
+ * cog.h
+ *
+ *  Created on: Mar 25, 2018
+ *      Author: itto
+ */
+
+#ifndef LIB_DESCRIPTING_COG_H_
```
Fluci: A more modern variant to header guards would be `#pragma once` (see other header files) But I don't really care.

lib/matching/nearest_neighbour.h
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+#ifndef LIB_MATCHING_NEAREST_NEIGHBOUR_H_
+#define LIB_MATCHING_NEAREST_NEIGHBOUR_H_
+
+#include "matching.h"
+
+namespace MouseTrack {
+
+class NearestNeighbour:Matching {
```
Fluci: You probably need `public Matching` (see [cpp Inheritance](https://www.tutorialspoint.com/cplusplus/cpp_inheritance.htm), section `Type of inheritance`).

lib/descripting/cog.h
```diff
@@ -0,0 +1,25 @@
...
+#define LIB_DESCRIPTING_COG_H_
+
+#include "descripting.h"
+
+namespace MouseTrack {
+
+class CenterOfGravity:Descripting {
```
Fluci: You probably need `public Matching` (see [cpp Inheritance](https://www.tutorialspoint.com/cplusplus/cpp_inheritance.htm), section `Type of inheritance`).

lib/matching/nearest_neighbour.cpp
```diff
@@ -0,0 +1,16 @@
...
+ *
+ *  Created on: Mar 25, 2018
+ *      Author: itto
+ */
+
+#include "nearest_neighbour.h"
+
+namespace MouseTrack {
+
+std::vector<long> NearestNeighbour::operator()(const std::vector<std::shared_ptr<const ClusterDescriptor>>& descriptors, const std::vector<ClusterChain>& chains) {};
```
Fluci: Somthing like `{return std::vector<long>{};}` would be neat, otherwise it will crush at runtime depending on the compiler settings

lib/descripting/cog.cpp
```diff
@@ -0,0 +1,17 @@
...
+ *  Created on: Mar 25, 2018
+ *      Author: itto
+ */
+
+#include "cog.h"
+
+namespace MouseTrack {
+
+/// Takes a cluster and its referencing PointCloud and creates a cluster descriptor
+std::unique_ptr<ClusterDescriptor> CenterOfGravity::operator()(const Cluster& cluster, const PointCloud& cloud) const {};
```
Fluci: Somthing like `{return std::unique_ptr<ClusterDescriptor>(...);}` would be neat, where `...` is an existing descriptor (otherwise it will crush at runtime depending on the compiler settings). Feel free to create a `CogDescriptor` (or so) inheriting from ClusterDescriptor.

Fluci (commented):
Like it, just some minor corrections

itko:
Fixed all except last one. There must be some sort of tool to automatically create the classes and member declarations/definitions of an implemented class given an abstract class, no?

This would reduce a lot of development overhead. Especially once we start trying out different algorithms. E.g. it would reduce the time needed to create the files and classes for every new implementation of Descripting.

A n automated script for this must exist!

lib/descripting/cog.cpp
```diff
@@ -0,0 +1,17 @@
+/* \file
```
Fluci: Doxygen is very picky in this regard (I've tried to change the behavior in the past, never worked), it expects two stars for the first line `/** \file`. Otherwise, it ignores the comment.

lib/descripting/cog.h
```diff
@@ -0,0 +1,21 @@
+/* \file
```
Fluci: `/**`

lib/matching/nearest_neighbour.cpp
```diff
@@ -0,0 +1,18 @@
+/* \file
```
Fluci: `/**`

lib/matching/nearest_neighbour.h
```diff
@@ -0,0 +1,24 @@
+/* \file
```
Fluci: `/**`

Fluci (changes requested):

Fluci:
Regarding the mocking of classes: As discussed, this is the IDE's job, I'm sure there's a button for it. Sometimes it's also called "stubs", mocking is mostly used in the context of unit tests, as far as I can tell.

itko:
Take a look. I'm new to smart pointers so I hope it makes sense. 

**Question**: if the Compare method of ClusterDescriptor takes in another ClusterDescriptor as input, doesn't this allow the possibility of comparing two different types of descriptors to each other?

Fluci:
Looks good to me now. 

Yes, it does allow to compare two different subclasses of ClusterDescriptor. Whether this makes sense, depends on the application. Since we probably won't mix different subclasses, I would just perform a dynamic down cast and let it fail for non-equal classes:

``` C++
Center* otherCenter = std::dynamic_cast<otherCenter*>(other);
```

PR #7, opened 2018-03-26
feature/lh/mocks

luhug:
I've written a test case which gives no errors so the basic functionality is given.
The mock clustering algorithm just assigns points to clusters randomly based on the number of clusters k you give it (also explained in documentation)

### DIFF
--- a/lib/clustering/mean_shift.cpp
+++ b/lib/clustering/mean_shift.cpp
@@ -5,17 +5,22 @@
 ///
 
 
-#include "random_clustering.h"
+#include "mean_shift.h"
 #include <stdlib.h>
+#include <time.h>
+#include <algorithm>
 namespace MouseTrack {
 
-RandomClustering::RandomClustering(int k) : _k(k) {
+MeanShift::MeanShift(int k) : _k(k) {
     //empty
 }
 
-std::vector<Cluster> RandomClustering::operator()(const PointCloud& cloud) const {
+std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
 
 	std::vector<Cluster> clustering = std::vector<Cluster>(_k);
+
+  //Initialize RNG
+	srand(time(NULL));
 
 	//Assign random cluster to each point in cloud
 	for(PointIndex i=0; i<cloud.size(); i++) {
@@ -25,9 +30,14 @@ std::vector<Cluster> RandomClustering::operator()(const PointCloud& cloud) const
 	}
 
 	//get rid of empty clusters
-	for (int i=0; i<clustering.size(); i++) {
-		if (clustering[i].points().size() == 0) clustering.erase(clustering.begin() + i);
-	}
+	int nextInsert = 0;
+  for(int i = 0; i < clustering.size(); i++){
+    if(!clustering[i].points().empty()){
+      clustering[nextInsert] = clustering[i];
+      nextInsert++;
+    }
+  }
+	clustering.erase(clustering.begin() + nextInsert, clustering.end());
 
 	return clustering;
 }
--- a/lib/clustering/mean_shift.cpp
+++ b/lib/clustering/mean_shift.cpp
@@ -8,7 +8,6 @@
 #include "mean_shift.h"
 #include <stdlib.h>
 #include <time.h>
-#include <algorithm>
 namespace MouseTrack {
 
 MeanShift::MeanShift(int k) : _k(k) {
--- a/lib/clustering/random_clustering.cpp
+++ b/lib/clustering/random_clustering.cpp
@@ -5,24 +5,28 @@
 ///
 
 
-#include "randomClustering.h"
+#include "random_clustering.h"
 #include <stdlib.h>
-
 namespace MouseTrack {
 
 RandomClustering::RandomClustering(int k) : _k(k) {
     //empty
 }
 
-std::vector<Cluster> RandomClustering::operator()(const PointCloud& cloud) {
+std::vector<Cluster> RandomClustering::operator()(const PointCloud& cloud) const {
 
 	std::vector<Cluster> clustering = std::vector<Cluster>(_k);
-	
 
+	//Assign random cluster to each point in cloud
 	for(PointIndex i=0; i<cloud.size(); i++) {
 		//Assign index i to random cluster in clustering
 		int cluster = rand() % _k;
-		clustering[i].points().push_back(i);
+		clustering[cluster].points().push_back(i);
+	}
+
+	//get rid of empty clusters
+	for (int i=0; i<clustering.size(); i++) {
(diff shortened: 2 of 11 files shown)


lib/clustering/random_clustering.cpp
```diff
@@ -0,0 +1,35 @@
...
+	//Assign random cluster to each point in cloud
+	for(PointIndex i=0; i<cloud.size(); i++) {
+		//Assign index i to random cluster in clustering
+		int cluster = rand() % _k;
+		clustering[cluster].points().push_back(i);
+	}
+
+	//get rid of empty clusters
+	for (int i=0; i<clustering.size(); i++) {
+		if (clustering[i].points().size() == 0) clustering.erase(clustering.begin() + i);
```
Fluci: It's ok, don't correct it, since it's just a mock and empty clusters are ok. 

But since I like to think about algorithms, there's probably a bug: 

Assume a vector `clustering` with clusters of size [0,0,1].

for i = 0: clustering := [0, 1]

for i = 1: nothing changes

returns [0, 1]

This means: if an element is deleted, the following element is moved to its location, but `i` is not updated, hence the moved element is never checked. -> you need to `i--` when erasing.

If you go backwards `for(int i = clustering.size() -1 ; 0 <= i; i--)`, this line actually works.

BTW: this loop is "slow" (total of O(n^2)) since every erase can cause a reallocation of all elements behind `i`: [docs](http://www.cplusplus.com/reference/vector/vector/erase/) 

I think the fastest version, would be

``` C++
int nextInsert = 0;
for(int i = 0; i < clustering.size(); i++){
    if(!clustering[i].points().empty()){
      clustering[nextInsert] = clustering[i];
      nextInsert++;
}
}

```

But there's probably a library function for this (`std::copy_if`?)

lib/clustering/random_clustering.test.cc
```diff
@@ -0,0 +1,44 @@
...
+
+namespace utf = boost::unit_test;
+
+
+BOOST_AUTO_TEST_CASE( random_clustering_read_write ) {
+    MouseTrack::PointCloud pc;
+    const int pcsize = 8;
+    pc.resize(pcsize);
+
+    for (int i=0; i<8; i++) {
```
Fluci: I'm not against magic numbers in unit tests, but since you define `pcsize` anyway, I think you should use it too:

```
for(int i = 0; i < pcsize; i += 1) {
```

lib/clustering/random_clustering.cpp
```diff
@@ -0,0 +1,35 @@
...
+}
+
+std::vector<Cluster> RandomClustering::operator()(const PointCloud& cloud) const {
+
+	std::vector<Cluster> clustering = std::vector<Cluster>(_k);
+
+	//Assign random cluster to each point in cloud
+	for(PointIndex i=0; i<cloud.size(); i++) {
+		//Assign index i to random cluster in clustering
+		int cluster = rand() % _k;
```
Fluci: You might want to avoid the usage of `rand()` as much as possible without a seed value. Assume there are cases where `rand` can produce invalid input. Since it is random, this only happens sometimes and you have no clue, what pattern causes the problem. To use a seed, see [srand](http://www.cplusplus.com/reference/cstdlib/srand/).

But since this is a mock, it's ok.

lib/generic/cluster.h
```diff
@@ -15,6 +15,9 @@ namespace MouseTrack {
 /// The corresponding point cloud is not linked in this data structure.
 class Cluster {
 public:
+    /// Construct empty cluster object
+    Cluster();
```
Fluci, line 19: Thanks, I forgot this one. :)

Fluci (changes requested):
Very good, I like unit tested code. 

I think, I formulated my request wrongly, really sorry about that. 

What I expected:

- The skeleton of the class(es) you intend to write. Something like `mean_shift.cpp` or `k_means.cpp` where the methods don't have to do anything (at the moment) except to return valid values (empty vector in this case). This is called stubbing/mocking and is a common technique to decouple architecture design from implementation.

Could you rename the files and classes according to your algorithm? You don't need to remove the existing functionality, since it's actually cool that it returns something non empty. You can replace it later with your implementation of mean shift.

BTW: I still reviewed the algorithms (I like thinking about code), but you don't have to care about it.

luhug:
@Fluci I renamed the files. I also applied the other changes you suggested anyway as I want to learn to write good code :) 

Your faster method to get rid of empty clusters was missing something. You have to erase the "obsolete" clusters at the end (bc you just copy them to lower indices). I added the following after your snippet:
`clustering.erase(clustering.begin() + nextInsert, clustering.end());`

For a random seed I hope it's ok to use the time(?)

lib/generic/cluster.h
itko: Is this necessary?

itko:
In the future make the title of the pull reuqest a little more descriptive!

lib/generic/cluster.h
luhug: Yes, for example if you want to create an empty cluster object and fill it with points afterwards, or if you initialize a std::vector of clusters it calls the empty constructor on every element.

lib/clustering/mean_shift.cpp
```diff
@@ -0,0 +1,44 @@
...
+MeanShift::MeanShift(int k) : _k(k) {
+    //empty
+}
+
+std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
+
+	std::vector<Cluster> clustering = std::vector<Cluster>(_k);
+
+  //Initialize RNG
+	srand(time(NULL));
```
Fluci, line 22: It really depends, what you want to use it for. If you generate passwords, this is a very good idea.
If you write physical simulations or run an algorithm, you want more control/a fixed seed (choose your favorite integer). 

It's all about repeatability: If you use something "random" like time, debugging might get really tricky, since you don't know the seed that causes a certain bug. This is desired in cryptographic applications (like password encryption or so). For nearly everything else, having this non-deterministic component, can cause a lot of problems.

lib/clustering/mean_shift.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+	//get rid of empty clusters
+	int nextInsert = 0;
+  for(int i = 0; i < clustering.size(); i++){
+    if(!clustering[i].points().empty()){
+      clustering[nextInsert] = clustering[i];
+      nextInsert++;
+    }
+  }
+	clustering.erase(clustering.begin() + nextInsert, clustering.end());
```
Fluci, line 39: You're right, totally forgot about this part X)

lib/generic/cluster.h
Fluci: :+1:

Fluci (approved):

PR #8, opened 2018-03-27
MatlabReader: reads the files exported by the matlab script

Fluci:
This branch got slightly larger than I expected, but here the most important changes:

- new dependency to read pngs: `$ sudo apt install libpng16-dev` (libpng12-dev should also be fine)
- Adjusted travis configuration to support libpng and C++11 regex
- Added section "Versions requirements" in the README, since testing with travis taught me a lot about which versions work with our stuff. We should add issues we encounter in the future.
- MatlabReader: 
  - Takes a path to a directory to which the matlab script exported a bag file. 
  - Gives easy access to  the file data: All you need to do is calling the corresponding method, the object knows about expected files and how to read them. 
  - It gains this knowledge from hard-coded file names and a preflight, where it works out the frame and stream index ranges (and a little bit more).
  - Only two unit tests which make sure, a well formatted directory succeeds to be read.
- generic functions to read pngs
- generic functions to read csvs
- Testing reading/writing is a pain with unit-testing, I hope the hard-coded paths aren't to fragile. There's a `./run_tests.sh` script which takes care of everything, just run it in your build folder.


README.md
```diff
@@ -112,3 +126,4 @@ BOOST_LOG_TRIVIAL(trace) << "A trace severity message";
     This is an error with fatal consequences, it might cost a lot of money/lives/cute kittens if ignored.
 
     Example: A safety check in a financial application trading stocks failed, million of dollars could be lost.
+
```
Fluci, line 129: Add section about testing script: how to run tests

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+        firstFrame = std::min(firstFrame, f);
+        lastFrame = std::max(lastFrame, f);
+    }
+    void addSuffix(std::string suff) {
+        suffixes.insert(suff);
+    }
+};
+
+// base file names for bag files
+const std::string NORMALIZED_DISPARITY_KEY{"depth_normalized"};
```
Fluci, line 66: () vs {} consistency

Fluci: [Prefer {} over () for initalization](https://stackoverflow.com/questions/18222926/why-is-list-initialization-using-curly-braces-better-than-the-alternatives)

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+const std::string REFERENCE_PICTURE_KEY("pic");
+const std::string ROTATION_CORRECTION_KEY("params_R");
+const std::string CAMERA_CHAIN_KEY("params_camchain");
+
+const std::set<std::string> EXPECTED_CHANNELS {NORMALIZED_DISPARITY_KEY, RAW_DISPARITY_KEY, FRAME_PARAMETERS_KEY, REFERENCE_PICTURE_KEY};
+const std::set<std::string> EXPECTED_FILES {ROTATION_CORRECTION_KEY, CAMERA_CHAIN_KEY};
+
+MatlabReader::MatlabReader(fs::path root_directory) : _root(fs::absolute(root_directory)), _valid(false) {
+    if(!fs::exists(_root)){
+        std::stringstream ss;
```
Fluci: Logging + throwing

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+        BOOST_LOG_TRIVIAL(trace) << "Found path: " << path.string();
+        int evaluatedSymlinks = 0;
+        while(evaluatedSymlinks < _followSymlinkDepth && fs::is_symlink(path)){
+            path = fs::read_symlink(path);
+            BOOST_LOG_TRIVIAL(trace) << "Followed symbolic link to: " << path.string();
+            evaluatedSymlinks += 1;
+        }
+        if(!fs::is_regular_file(path)){
+            // add original path to ignore list
+            BOOST_LOG_TRIVIAL(trace) << "Ignored non regular file: " << p.path().string();
```
Fluci: Follow README convention

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+    // TODO: Boost 1.54 does not support fs::directory_iterator as we would like to use it here
+    // for(const auto& p : fs::directory_iterator(_root)){
+    // Later versions do, but I can't make travis use the newer versions, hence this older code until an update is possible
+    for(const auto& p : boost::make_iterator_range(fs::directory_iterator(_root), fs::directory_iterator())){
+        fs::path path = p.path();
+        BOOST_LOG_TRIVIAL(trace) << "Found path: " << path.string();
+        int evaluatedSymlinks = 0;
+        while(evaluatedSymlinks < _followSymlinkDepth && fs::is_symlink(path)){
+            path = fs::read_symlink(path);
+            BOOST_LOG_TRIVIAL(trace) << "Followed symbolic link to: " << path.string();
```
Fluci: info or debug level?

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+    /// expresson to extract the numbers. First we always widen
+    /// the index range on a per channel basis.
+    /// Then we take the smallest common range as final result.
+    /// While iterating over all entries, we also store the file paths to the files
+    /// for easy look up later on.
+    /// We list the paths from the directory iterator in the ignored file list, but
+    /// store the paths from following symlinks in all other cases.
+    std::unordered_map<std::string, IndexAggregateSF> aggSF;
+    std::unordered_map<std::string, IndexAggregateF> aggF;
+    std::regex nameShape("^([A-Za-z0-9_-]+?)(?:_s_([0-9]+))?_f_([0-9]+)\\.([A-Za-z0-9]+)$", std::regex::ECMAScript);
```
Fluci: Any point int moving this definition to a more static context?

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+            channel = cm[1];
+            stream = std::stoi(cm[2]);
+            frame = std::stoi(cm[3]);
+
+            auto& a = aggSF[channel];
+            a.addSuffix(suffix);
+            a.addStream(stream);
+            a.addFrame(frame);
+        } else {
+            BOOST_LOG_TRIVIAL(trace) << "Ignoring path " << p.path().string() << ", does not match any expected file name.";
```
Fluci: info level

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+    for(const auto& kv : aggF){
+        if(EXPECTED_CHANNELS.find(kv.first) == EXPECTED_CHANNELS.end()){
+            _ignoredChannels.push_back(kv.first);
+        }
+        firstFrame = std::min(firstFrame, kv.second.firstFrame);
+        lastFrame = std::max(lastFrame, kv.second.lastFrame);
+    }
+
+    // TODO: check file formats
```
Fluci: TODO

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+const std::vector<fs::path>& MatlabReader::ignoredPaths() const {
+    return _ignoredPaths;
+}
+
+const std::vector<std::string>& MatlabReader::ignoredChannels() const {
+    return _ignoredChannels;
+}
+
+std::vector<Frame> MatlabReader::frames(FrameNumber f) const {
```
Fluci: test this method

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+    return read4x4MatrixList(p);
+}
+
+
+std::vector<Eigen::Matrix4d> MatlabReader::read4x4MatrixList(const fs::path& file) const {
+    const auto fileContent = read_csv(file.string());
+    // matlab reshapes matrices colum wise: [1 2; 3 4] -> reshape -> [1 3 2 4]
+    std::vector<Eigen::Matrix4d> result(fileContent.size());
+    for(size_t mat = 0; mat < fileContent.size(); mat += 1){
+        for(size_t j = 0; j < 4; j += 1){
```
Fluci: no unnamed magic numbers

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+        }
+        if(!fs::is_regular_file(c)){
+            continue;
+        }
+        return c;
+    }
+    throw "no path found.";
+}
+
+std::string MatlabReader::createFilename(const std::string& base, StreamNumber s, FrameNumber f, const std::string& suffix) const {
```
Fluci: unused

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+    std::stringstream ss;
+    ss << base;
+    ss << "_f_" << f;
+    ss << "_s_" << s;
+    ss << "." << suffix;
+    return ss.str();
+}
+
+std::string MatlabReader::createFilename(const std::string& base, FrameNumber f, const std::string& suffix) const {
```
Fluci: unused

app/matlab_reader.cpp
```diff
@@ -0,0 +1,452 @@
...
+    for(size_t i = 0; i < fileContent.size(); i += 1){
+        for(size_t j = 0; j < fileContent[i].size(); j += 1){
+            double entry = std::stod(fileContent[i][j]);
+            result(i,j) = entry;
+        }
+    }
+    return result;
+}
+
+// helpers
```
Fluci: remove

app/matlab_reader.h
```diff
@@ -0,0 +1,170 @@
...
+        bool operator<(const ElementKey& other) const;
+    };
+    /// Collect data relevant paths in an organized way
+    struct FilePaths {
+        std::map<ElementKey, std::set<fs::path>> frames;
+        std::set<fs::path> rotationCorrectionsPath;
+        std::set<fs::path> camchainPath;
+    };
+
+    fs::path _root;
```
Fluci: comment missing

app/matlab_reader.h
```diff
@@ -0,0 +1,170 @@
...
+
+    /// We collect expected paths here
+    FilePaths _files;
+
+    /// Some files are expected to be accessed many times, we cache them here
+    ConstParameters _cache;
+    /// number of symlinks to evaluate (0: don't follow symlinks, 1: evaluate one link, etc.)
+    int _followSymlinkDepth = 1000;
+
+    void preflight();
```
Fluci: Comment

app/matlab_reader.h
```diff
@@ -0,0 +1,170 @@
...
+    std::vector<Eigen::Matrix4d> read4x4MatrixList(const fs::path& file) const;
+
+    /// Reads a csv file into one matrix
+    Eigen::MatrixXd readMatrix(const fs::path& file) const;
+
+    /// constructs a filename according to the expected pattern
+    std::string createFilename(const std::string& base, StreamNumber s, FrameNumber f, const std::string& suffix) const;
+
+    /// as the other createFilename but without stream index
+    std::string createFilename(const std::string& base, FrameNumber f, const std::string& suffix) const;
```
Fluci: remove

app/matlab_reader.test/README.md
```diff
@@ -0,0 +1,3 @@
+This directory holds mostly empty files that imitate the file structure of a real input folder for MatlabReader.
+
+Most files are empty, since their content is irrelevant for testing.
```
Fluci: not empty

lib/generic/frame.h
```diff
@@ -0,0 +1,30 @@
...
+#include "generic/disparity_map.h"
+#include <Eigen/Core>
+
+namespace MouseTrack {
+
+
+/// Holds all information related to a single frame of a stream
+struct Frame {
+    DisparityMap normalizedDepth;
+    DisparityMap rawDepth;
```
Fluci: rename

lib/generic/read_png.cpp
```diff
@@ -0,0 +1,105 @@
...
+        // OK
+        break;
+    case PNG_COLOR_TYPE_GRAY_ALPHA:
+        // OK
+        break;
+    case PNG_COLOR_TYPE_PALETTE:
+        fclose(fp);
+        throw "Palette pngs are not supported";
+    case PNG_COLOR_TYPE_RGB:
+        // TODO: Log: we ignore GB
```
Fluci: TODO

app/matlab_reader.cpp
Fluci: done

README.md
Fluci: done

app/matlab_reader.cpp
Fluci: done

app/matlab_reader.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+const std::vector<fs::path>& MatlabReader::ignoredPaths() const {
+    return _ignoredPaths;
+}
+
+const std::vector<std::string>& MatlabReader::ignoredChannels() const {
+    return _ignoredChannels;
+}
+
+std::vector<Frame> MatlabReader::frames(FrameNumber f) const {
```
Fluci: can this method throw? should probably have a `nothrow`, maybe a valid flag on Frame?

app/matlab_reader.h
```diff
@@ -0,0 +1,166 @@
...
+    Picture picture(StreamNumber s, FrameNumber f) const;
+
+    /// Fetch rotation corrections for cameras
+    std::vector<Eigen::Matrix4d> rotationCorrections() const;
+
+    /// Fetch the camera chain (8 rotation matrices that define how the cameras are positioned to each other)
+    std::vector<Eigen::Matrix4d> camchain() const;
+
+    /// Fetch all data belonging to frame f of and all streams
+    std::vector<Frame> frames(FrameNumber f) const;
```
Fluci: This return type is probably wrong.

- There are things depending on the frame and the stream.
- There are things depending on the frame index (equal for all frames with same index)
- There are things depending on the stream.
- There are things that depend on nothing. 

We do want to denormalize a bit. We probably need a construct like a "FrameStack" or "FrameWindow" which holds parameters constant for a frameIndex and also a list of "Frame" with all parameters differing between frames.

We might need to define the terminology at some point.

Fluci: I think I should go to bed. If we assume a camchain to be defined according to the stream indexes, this actually should work, since each Frame holds one camchain matrix and one rotation corrections. I'll think about it tomorrow.

Fluci: I've made my mind up, to be consistent with other concepts like "Cluster" an object like "FrameWindow" makes sense. It also gives the flexibility to add values constant accross frames with the same index.

lib/generic/read_png.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+#include <png.h>
+#include <boost/log/trivial.hpp>
+
+namespace MouseTrack {
+
+Picture read_png_normalized(const std::string& path){
+    auto mat = read_png(path);
+    Picture result = mat.cast<double>();
+    result /= 256.0;
```
luhug: Isn't the max pixel value 255, so we should divide by 255?

Fluci: you're right, I'll adjust it.

luhug (approved):

itko (approved):
Looked through quickly. In the future would try to segment this pull request so it's less to review all at once. Opinion?

Do we really need to make our own csv reader? Aren't there quite a few open-source ones we can use?

Will we need csv writing capabilities in the future (e.g. when outputting the mouse position)? If so, should combine with the read_csv library

Approving anyways. Seems good.

Fluci:
**Segmenting pull requests:**

In theory, that's a good idea and I know, that for example the Linux guys do everything as "atomic" as possible. They've denied patches from AMD in the past, because they were just large blobs. 

I've tried that in the past with half a dozen pull requests in parallel. Problems appear, when the reviews take a long time. Things tend to diverge while fixing stuff. This is not that bad, if things don't depend on each other (like the csv and png reader). But it can get very annoying, when they touch the same files (CMakeLists, interfaces, etc.). 

You also get a temporal ordering. When MatlabReader depends on csv and png, those two pull requests first need to be accepted into master, otherwise MatlabReader will show the changes of csv and png too while reviewing. I leave open, whether that's a good thing.

I haven't figured out how to merge large batches of code reasonably, so I'm very glad for new suggestions. 

I usually go with the rule: if I've created it together, I'l push it together. But if it's an "orthogonal" addition, I make a separate push (as done with the interface change of TrajectoryBuilder).

**CSV Dependency**

I've checked, if it is worth using a csv-dependency, there is one available via apt (libcsv). I decided against it. My reasons are:

- I want working and maintained dependencies. The project link in `apt` points to source forge, while the official repository changed to github, which in turn isn't even properly versioned and most files haven't been touched in over a year. This doesn't look trustworthy at the moment.
- Adding another csv library, not supported by apt, is basically a no-go for this project, unless it's absolutely necessary. Compiling C++ dependencies (and their dependencies) from source, can take a lot of time and break very easily.
- C API's are very annoying.
- Reading/Writing csv with numbers is easy. (Reading CSV with text is very hard!) That's all we need for this project.

I'm open to switsch to libcsv in the future, should the need appear.

Yes, libpng is also a C API and there exists a libpng++ building on top of it. Why did I choose libpng and not libpng++? We already use libpng, it's a dependency of OpenCV, so I effectively avoid a new dependency. Checking their [website](http://savannah.nongnu.org/projects/pngpp/) revealed the youngest news entry to be from 2015 and the last still open bug report from mid 2017. This also didn't pass my "robustness-filter".

PR #11, opened 2018-03-30
Qt hello world added

Fluci:
**Outline**

I'm using a model-view-controller approach. This means:
- We have a model object (Pipeline) that is responsible for the core-logic, it coordinates the processing of frames.
- We have a view (Qt), responsible to display stuff to the user.
- We have a controller (QtController, Luzian's playground), responsible to process events (user clicks, new data from pipeline), connecting view and model. 

The pipeline implementation will follow in a later pull request, together with an observer pattern, allowing for communication with the controller.

Note: travis doesn't work yet with qt, and I would like to skip it for the moment (debugging travis is very time consuming).

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -3,7 +3,14 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
 # list here all source files appart from main
 set(source_files
+        controller.cpp
         matlab_reader.cpp
+        pipeline.cpp
+        )
+
+# list here all source files relating with the gui
+set(gui_source_files
+        qt_controller.cpp
         )
 
 # list here your testing files (*.test.cc) of your module
@@ -26,6 +33,22 @@ find_package( Boost COMPONENTS
 
 ADD_DEFINITIONS(-DBOOST_LOG_DYN_LINK)
 
+if(ENABLE_GUI)
+  find_package(Qt5Core)
+  
+  # Find the QtWidgets library
+  find_package(Qt5Widgets)
+  set(GUI_LIBRARIES
+    Qt5::Widgets
+    )
+  add_definitions(-DENABLE_GUI)
+else(ENABLE_GUI)
+    # reset gui sources
+    set(gui_source_files "")
+endif(ENABLE_GUI)
+
+# Use the Widgets module from Qt 5.
+
 # compile everything into a library and then link into the target executable (tests, main, etc.)
 # this should increase compilation speed
 add_library(libmousetrackAll ${source_files})
@@ -39,13 +62,23 @@ target_link_libraries(libmousetrackAll
     ${CMAKE_THREAD_LIBS_INIT}
 )
 
+
 # create the mousetrack executable
-add_executable(mousetrack main.cpp)
-target_link_libraries(mousetrack libmousetrackAll)
+add_executable(mousetrack main.cpp ${gui_source_files})
+target_link_libraries(mousetrack
+    libmousetrackAll
+    ${GUI_LIBRARIES}
+    )
 
 # create unit test executable
 add_executable(appTests ${test_files})
 target_link_libraries(appTests
     libmousetrackAll
     ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY}
-)
+    )
+
+
+# create a target for the headers, we don't actually need the result
+# but this way, IDE's also show header files in their project overviews
+file(GLOB_RECURSE appHeader_files "*.h")
+add_custom_target(appHeaders SOURCES ${appHeader_files})
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -8,7 +8,7 @@ set(source_files
         pipeline.cpp
         )
 
-# list here all source files relating with the gui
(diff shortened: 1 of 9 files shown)


app/CMakeLists.txt
```diff
@@ -26,6 +33,22 @@ find_package( Boost COMPONENTS
...
+  set(GUI_LIBRARIES
+    Qt5::Widgets
+    )
+  add_definitions(-DENABLE_GUI)
+else(ENABLE_GUI)
+    # reset gui sources
+    set(gui_source_files "")
+endif(ENABLE_GUI)
+
+# Use the Widgets module from Qt 5.
```
Fluci: remove

app/CMakeLists.txt
```diff
@@ -3,7 +3,14 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
 # list here all source files appart from main
 set(source_files
+        controller.cpp
         matlab_reader.cpp
+        pipeline.cpp
+        )
+
+# list here all source files relating with the gui
```
Fluci: related

app/pipeline.h
```diff
@@ -0,0 +1,47 @@
...
+/// Note: This class is mostly mocked, an improved interface and implementation comes later.
+
+
+namespace MouseTrack {
+
+/// This object is responsible to coordinate the processing steps
+///
+/// - It knows how to acquire new frames
+/// - It knows how to push the frames through the pipeline steps
+/// - It informs other objects about (inbetween-)results via an observer pattern (PipelineObserver)
```
Fluci, line 18: comes with the next pull request, I'll leave it here

app/qt_controller.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+    // set up application
+    QApplication app (argc, argv);
+
+    // show a button in a window
+    QPushButton button ("Hello world !");
+    button.setToolTip("A tooltip");
+    button.show();
+    // start pipeline
+    pipeline().start();
```
Fluci, line 23: @luhug : the entire method is just a draft, feel free to move stuff around as it is convenient for you.

app/controller.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include "pipeline.h"
+
+namespace MouseTrack {
+
+class Controller {
+public:
+
+    /// Called by main after creation
+    virtual int main(int argc, char *argv[]) = 0;
```
Fluci, line 16: This might be exchanged by an object resulting from parsing the command line, don't use it for the moment

app/pipeline.h
itko: What does join() do? I'm not sure it's explained clearly in the code.

app/qt_controller.cpp
itko: What's controller.cpp and what is qt_controller.cpp?

itko (commented):
Again, I just looked through the files to make sure structure is understandable. I trust it works fine.

app/pipeline.h
Fluci: Oh, you're right, I totally forgot to mention the entire worker thread part. I'll put a note.

app/qt_controller.cpp
Fluci: `Controller` defines an abstract class, controller.cpp implements the interface for the pipeline access. Each controller is responsible to connect the ui with the model code (the part that does the actual work). I've decided to put the pipeline as a member into the controller class, since every controller will interact with a pipeline, at least in our case. QtController is a child of controller. It knows how to create a qt gui and interact with the user. I've also written a cliController that allows us to perform integration tests (comes with a later push).

app/pipeline.h
Fluci: Done, I hope it's now a bit clearer. We only need a worker thread, because we're working with a gui. We need one thread updating the views and interacting with the user. And we need one other to do the actual work. Doing everything in the run loop of qt could be possible, but would cause our interface to freeze terribly as soon as we do some processing.

luhug (approved):
Looks good, thanks Felice!

app/pipeline.h
```diff
@@ -16,6 +16,12 @@ namespace MouseTrack {
 /// - It knows how to acquire new frames
 /// - It knows how to push the frames through the pipeline steps
 /// - It informs other objects about (inbetween-)results via an observer pattern (PipelineObserver)
+///
+/// The pipeline runs one or multiple worker threads to process the frame stream concurrently.
+/// Start the processing with `start()`. If you want to terminate the pipeline early, call `stop()`,
+/// a signal will be sent to the pipeline, it should stop as soon as possible.
+/// Always call `join` before terminating, it will make sure that all worker threads terminated and delivered their data.
```
itko, line 23: Still not so sure what you mean. I guess the explanation makes sense. But the verb "join" doesn't seem to correspond to the explanation. Maybe I'm not understanding the concept itself.

Fluci: Don't know, I'm not a native speaker. The terminology comes from the world of concurrency, where join is the standard term to merge two execution paths into one (see std::thread::join() as an example)

PR #12, opened 2018-04-03
Registration added

Fluci:
Combines the images and disparity maps into a point cloud.

### DIFF
--- a/lib/registration/disparity_registration.h
+++ b/lib/registration/disparity_registration.h
@@ -1,0 +1,25 @@
+/// \file
+/// Maintainer: Felice Serena
+///
+
+#pragma once
+
+#include "registration.h"
+
+namespace MouseTrack {
+
+class DisparityRegistration : public Registration {
+public:
+    PointCloud operator()(const FrameWindow& window) const;
+private:
+    /// constant from fpga set up
+    double min_disparity = 1*2+32;
+    /// constant from fpga set up
+    int xshift = 22;
+    /// constant from fpga set up
+    int yshift = -8;
+    /// constant from fpga set up
+    int frame_border = 80;
+};
+
+} // MouseTrack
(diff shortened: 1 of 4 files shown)


lib/registration/disparity_registration.h
```diff
@@ -0,0 +1,25 @@
...
+    PointCloud operator()(const FrameWindow& window) const;
+private:
+    /// constant from fpga set up
+    double min_disparity = 1*2+32;
+    /// constant from fpga set up
+    int xshift = 22;
+    /// constant from fpga set up
+    int yshift = -8;
+    /// constant from fpga set up
+    int frame_border = 80;
```
luhug: Couldn't we make these four variables const? It's probably not important but just to be consistent.

luhug (commented):
I don't really understand all the math (Householder transformations etc.) but from my point of view, I can't spot any obvious issues. I guess it'll be easier to spot bugs once we have it integrated in our main process.

lib/registration/disparity_registration.h
Fluci: I actually tend to write accessors for them, since they are part of the set up, we/clients might want to change the values in the future.

Fluci:
Regarding bugs: It's mostly a C++ version of Dominik's script. Adding a Ply exporter is on my todo list, so we can view the points in meshlab or so for verification.

lib/registration/disparity_registration.h
Fluci: I probably also should adjust the description.

Fluci:
Added accessors. The values are still hard coded by default. I think I leave it like this and don't put them into a configuration file for the moment.

luhug (approved):

itko (approved):
Approved without close inspection. It's too much overhead to dig into this right now.

PR #13, opened 2018-04-03
Clean up and bug fixes for MatlabReader

Fluci:
- Added interface for MatlabReader
- fixed bug with regex cases
- catches for integer conversions
- allow to set frame range

### DIFF
--- a/lib/reader/reader.h
+++ b/lib/reader/reader.h
@@ -1,0 +1,32 @@
+/// \file
+/// Maintainer: Felice Serena
+///
+
+#pragma once
+
+#include "generic/types.h"
+#include "generic/disparity_map.h"
+#include "generic/frame_window.h"
+
+namespace MouseTrack {
+
+class Reader {
+public:
+
+    /// true if the reader is ready to provide output
+    virtual bool valid() const = 0;
+
+    /// index of first existing frame
+    virtual FrameNumber beginFrame() const = 0;
+
+    /// Index of first non existing frame at the end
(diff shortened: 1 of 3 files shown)


luhug (approved):
Looks good. Good work Felice! No way I could have coded the MatlabReader by myself

lib/reader/reader.h
```diff
@@ -0,0 +1,32 @@
...
+
+#pragma once
+
+#include "generic/types.h"
+#include "generic/disparity_map.h"
+#include "generic/frame_window.h"
+
+namespace MouseTrack {
+
+class Reader {
```
itko, line 13: What is the purpose of the reader interface? Are we going to use it outside MatLab readeR?

itko: And in this case, since you refer to frames, shouldn't it be something more descriptive instead of Reader? E.g.  "VideoReader"?

itko (approved):

lib/reader/reader.h
Fluci: There are things, the pipeline doesn't need to know about, but might be interesting for other parts of the application (fine grained access to file data). I want to provide these methods to Luzian, but don't want to use them accidentally in the pipeline. Plus, it is clean to have an interface for a part, we might want to replace some day (streaming for ROS, other sources ...). 
I'm not happy with the terminology either, but Frames aren't single images, but also everything associated with the frame, like camera parameters. I hope to generalize the interface some day, so we can read cached data like point clouds or cluster assignments.

PR #14, opened 2018-04-03
Implemented Mean Shift algorithm

luhug:
This is my implementation of the Mean Shift algorithm tailored to our purposes.
Some notes:

- I added an eigenVec() function to PointCloud::Point to easily convert a point in a point cloud to an Eigen::VectorXd type. This is useful/necessary to perform operations on the points. It is a VectorXd instead of Vector4d so that we can easily add more dimensions if needed. This is useful/necessary to perform operations on the points.
- The algorithm passes all the unit tests I wrote. Namely: read/write, single point, two points, three gaussian-distributed clusters of 100 points each.
- As far as I can tell, it clusters all of the above examples "correctly". Of course there may still be bugs, I guess we'll find out when we use it in the implementation ;)
- I added some boost log messages - @Fluci did I understand correctly how to use them?
- Concerning efficiency. It is probable that this is not the most efficient way to implement it as I am not at all an expert in this topic - at one point I have three nested for loops which all loop over a potentially large amount of elements. Please feel free to comment on how to make my algorithm more efficient.

### DIFF
--- a/lib/clustering/mean_shift.cpp
+++ b/lib/clustering/mean_shift.cpp
@@ -70,7 +70,7 @@ std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
 		}
 
 		if (clusters.size() != curr.size())
-			BOOST_LOG_TRIVIAL(warning) << "Cluster count does not equal mode count. (" << clusters.size() << " != " << curr.size();
+			BOOST_LOG_TRIVIAL(warning) << "Cluster count does not equal mode count. (" << clusters.size() << " != " << curr.size() << ")";
 
 		// Check convergence
 		if (prev.size() == curr.size()) {
--- a/lib/clustering/mean_shift.cpp
+++ b/lib/clustering/mean_shift.cpp
@@ -37,7 +37,6 @@ std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
 
 	// For each point...
 	for(PointIndex i = 0; i < clusters.size(); i++) {
-		prevCenter = points[i];
 		int iterations = 0; //for logging and abort condition
 
 		// ... iterate until convergence
--- a/lib/clustering/mean_shift.cpp
+++ b/lib/clustering/mean_shift.cpp
@@ -9,6 +9,7 @@
 #include <stdlib.h>
 #include <time.h>
 #include <math.h>
+#include <Eigen/Core>
 namespace MouseTrack {
 
 MeanShift::MeanShift(int k) : _k(k) {
@@ -42,17 +43,16 @@ std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
 	return clustering;
 }
 
-double MeanShift::euclidean_distance_squared(const Point& x, const Point& y) {
-	return 	(x.x() - y.x())*(x.x() - y.x()) +
-					(x.y() - y.y())*(x.y() - y.y()) +
-					(x.z() - y.z())*(x.z() - y.z()) +
-					(x.intensity() - y.intensity())*(x.intensity() - y.intensity());
+double MeanShift::apply_gaussian_kernel(const PointCloud& cloud, const PointIndex& point, const PointIndex& mean) {
+	const GaussianKernel kernel = {means[mean], _window_size};
+
+	// Euclidean distance squared btw. mean of kernel and point of interest
+	const double d = euclidean_distance_squared(means[mean], cloud[point]);
+	return exp(-d/(2*_window_size));
 }
 
-double MeanShift::apply_gaussian_kernel(const PointCloud& cloud, const PointIndex& point, const GaussianKernel& kernel) {
-	// Euclidean distance squared btw. mean of kernel and point of interest
-	const double d = euclidean_distance_squared(kernel.mean, cloud[point]);
-	return exp(-d/(2*kernel.variance));
+Eigen::Vector4d MeanShift::create_vector(const PointCloud& cloud, const PointIndex& index) {
+	return Eigen::Vector4d()
 }
 
 } //MouseTrack
--- a/lib/clustering/mean_shift.cpp
+++ b/lib/clustering/mean_shift.cpp
@@ -7,52 +7,36 @@
 
 #include "mean_shift.h"
 #include <stdlib.h>
-#include <time.h>
 #include <math.h>
-#include <Eigen/Core>
+#include <boost/log/trivial.hpp>
+#include <Eigen/Dense>
+
 namespace MouseTrack {
 
-MeanShift::MeanShift(int k) : _k(k) {
+MeanShift::MeanShift(double window_size) : _window_size(window_size) {
     //empty
 }
 
 std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
 
-	std::vector<Cluster> clustering = std::vector<Cluster>(_k);
-
-  //Initialize RNG
-	srand(time(NULL));
-
-	//Assign random cluster to each point in cloud
-	for(PointIndex i=0; i<cloud.size(); i++) {
-		//Assign index i to random cluster in clustering
-		int cluster = rand() % _k;
-		clustering[cluster].points().push_back(i);
-	}
-
-	//get rid of empty clusters
-	int nextInsert = 0;
-  for(int i = 0; i < clustering.size(); i++){
-    if(!clustering[i].points().empty()){
-      clustering[nextInsert] = clustering[i];
-      nextInsert++;
-    }
-  }
-	clustering.erase(clustering.begin() + nextInsert, clustering.end());
-
+	std::vector<Cluster> clustering = std::vector<Cluster>();
 	return clustering;
 }
 
-double MeanShift::apply_gaussian_kernel(const PointCloud& cloud, const PointIndex& point, const PointIndex& mean) {
-	const GaussianKernel kernel = {means[mean], _window_size};
+double MeanShift::apply_gaussian_kernel(const PointCloud& cloud, const PointIndex& idx, Eigen::MatrixXd mean) const {
+
+	// Convert to Eigen type and check if dimensions match
+	Eigen::MatrixXd point = cloud[idx].eigenVec();
+	if (mean.size() != point.size()) BOOST_LOG_TRIVIAL(error) << "Vector dimensions don't match";
 
 	// Euclidean distance squared btw. mean of kernel and point of interest
-	const double d = euclidean_distance_squared(means[mean], cloud[point]);
+	const double d = (mean-point).squaredNorm();
+
+	// Apply gaussian distribution
 	return exp(-d/(2*_window_size));
 }
 
-Eigen::Vector4d MeanShift::create_vector(const PointCloud& cloud, const PointIndex& index) {
-	return Eigen::Vector4d()
+void MeanShift::iterate() {
+	//TODO
 }
-
 } //MouseTrack
--- a/lib/clustering/mean_shift.cpp
+++ b/lib/clustering/mean_shift.cpp
@@ -10,7 +10,6 @@
 #include <math.h>
 #include <boost/log/trivial.hpp>
 #include <Eigen/Dense>
-#include <iostream>
 
 namespace MouseTrack {
 
@@ -19,11 +18,54 @@ MeanShift::MeanShift(double window_size) : _window_size(window_size) {
 }
 
 std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
+	// Covert point cloud to Eigen vectors
+	std::vector<Eigen::VectorXd> points;
+	for (int i=0; i<cloud.size(); i++) {
+		points.push_back(cloud[i].eigenVec());
+	}
+
+	std::vector<Eigen::VectorXd> prev(points.size());
+	std::vector<Eigen::VectorXd> curr = points;
+	bool has_converged = false;
+
+	while (!has_converged) {
+
+		// perform 1 iteration on each mode
+		for (int i=0; i<prev.size(); i++) {
+			curr[i] = iterate_mode(prev[i], prev);
+		}
+
+		// Merge step
+		// For every mode..
+		for (int i=0; i<curr.size(); i++) {
+			// Check modes we haven't already executed the (i)-loop for
+			for (int j=i+1; i<curr.size(); j++) {
+				Eigen::VectorXd diff = curr[i] - curr[j];
+				if (diff.norm() < MERGE_THRESHOLD) {
+					curr.erase(curr.begin() + j);
+					break;
+				}
+			}
+		}
+
+		// Check convergence
+		if (prev.size() == curr.size()) {
+			double total_movement;
+			for (int i=0; i < curr.size(); i++) {
+				total_movement += (curr[i]-prev[i]).norm();
+			}
+			if (total_movement <= CONVERGENCE_THRESHOLD) has_converged = true;
+		}
+		prev = curr;
+	}
+	BOOST_LOG_TRIVIAL(trace) << "MeanShift converged! #Clusters: " << curr.size();
+
+	// Build clusters
 	std::vector<Cluster> clustering = std::vector<Cluster>();
 	return clustering;
 }
 
-double MeanShift::apply_gaussian_kernel(const Eigen::MatrixXd point, const Eigen::MatrixXd mean) const {
+double MeanShift::apply_gaussian_kernel(const Eigen::VectorXd point, const Eigen::VectorXd mean) const {
 
 	// check if dimensions match
 	if (mean.size() != point.size()) BOOST_LOG_TRIVIAL(error) << "Vector dimensions don't match";
@@ -33,6 +75,19 @@ double MeanShift::apply_gaussian_kernel(const Eigen::MatrixXd point, const Eigen
 
 	// Apply gaussian distribution
 	return exp(-d/(2*_window_size));
+}
+
+Eigen::VectorXd MeanShift::iterate_mode(const Eigen::VectorXd mode, const std::vector<Eigen::VectorXd>& state) const {
+	//COG Normalization Factor
+	double normfact = 0;
+	//Rest of COG
+	Eigen::VectorXd cog = Eigen::VectorXd::Zero(mode.size());
+	for(const Eigen::VectorXd x : state) {
+		double temp = apply_gaussian_kernel(x,mode);
+		normfact += temp;
+		cog += temp * x;
+	}
+	return cog * (1.0/normfact);
 }
 
 } //MouseTrack
--- a/lib/clustering/mean_shift.cpp
+++ b/lib/clustering/mean_shift.cpp
@@ -61,7 +61,7 @@ std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
 			Eigen::VectorXd diff = currCenters[i] - currCenters[j];
 			if (diff.norm() < _merge_threshold) {
 				// Merge clusters. Erase one of the modes corresponding to the clusters and append points belonging to j to cluser of i
-				clusters[i].points().insert(clusters[i].points().begin(), clusters[j].points().begin(), clusters[j].points().end());
+				clusters[i].points().push_back(j);
 				clusters.erase(clusters.begin() + j);
 				currCenters.erase(currCenters.begin() + j);
 				// We need to decrement i here because there might be another cluster that wants to merge with i.
--- a/lib/clustering/mean_shift.cpp
+++ b/lib/clustering/mean_shift.cpp
@@ -58,8 +58,6 @@ std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
 
 				if (diff.norm() < MERGE_THRESHOLD) {
 					// Merge clusters. Erase one of the modes corresponding to the clusters and append points belonging to j to cluser of i
-					//for (PointIndex k : clusters[j].points()) {
-
 					for (int k=0; k < clusters[j].points().size(); k++) {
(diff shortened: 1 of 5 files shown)


lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
 
-MeanShift::MeanShift(int k) : _k(k) {
-    //empty
+MeanShift::MeanShift(double window_size) : _window_size(window_size) {
+	//empty
 }
 
 std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
+	BOOST_LOG_TRIVIAL(trace) << "MeanShift algorithm started";
+	// Covert point cloud to Eigen vectors
```
Fluci: typo: Convert

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
+	std::vector<Eigen::VectorXd> prev = points;
+	std::vector<Eigen::VectorXd> curr(points.size());
+	bool has_converged = false;
+  int iterations = 0; //for logging
+
+	// The MeanShift loop
+	while (!has_converged) {
+		iterations++;
+		if (iterations > MAX_ITERATIONS) {
+			BOOST_LOG_TRIVIAL(error) << "MeanShift: max number of iterations exceeded without convergence";
```
Fluci: I tend to go with an exception/some kind of error handling here, a simple log seems too weak in my opinion. Why? Not-to-converge is a valid result for this algorithm, as it is for every iterative algorithm, as long as we give finite resources. We probably want to inform the client in a more explicit way. 
Would a non-converged result be useful?
Maybe that's an argument for accessors to MAX_ITERATIONS. If the algorithm doesn't converge, it throws an exception and it is the client's responsibility to run the algorithm with more resources (higher MAX_ITERATIONS).

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
+		if (prev.size() == curr.size()) {
+			double total_movement = 0;
+			for (int i=0; i < curr.size(); i++) {
+				total_movement += (curr[i]-prev[i]).norm();
+			}
+			if (total_movement <= CONVERGENCE_THRESHOLD) has_converged = true;
+		}
+		prev = curr;
 	}
+	BOOST_LOG_TRIVIAL(trace) << "MeanShift converged! #Clusters: " << curr.size() << " #Iterations: " << iterations;
```
Fluci: :+1:

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
+			// Check modes we haven't already executed the (i)-loop for
+			for (int j=i+1; j<curr.size(); j++) {
+				Eigen::VectorXd diff = curr[i] - curr[j];
+
+				if (diff.norm() < MERGE_THRESHOLD) {
+					// Merge clusters. Erase one of the modes corresponding to the clusters and append points belonging to j to cluser of i
+					for (int k=0; k < clusters[j].points().size(); k++) {
+						clusters[i].points().push_back(clusters[j].points()[k]);
+					}
+					//BOOST_LOG_TRIVIAL(debug) << "Merging cluster " << j << " into cluster " << i << ", distance " << diff.norm();
```
Fluci: Either use or remove.

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
+		// Merge step
+		// For every mode..
+		for (int i=0; i<curr.size(); i++) {
+			// Check modes we haven't already executed the (i)-loop for
+			for (int j=i+1; j<curr.size(); j++) {
+				Eigen::VectorXd diff = curr[i] - curr[j];
+
+				if (diff.norm() < MERGE_THRESHOLD) {
+					// Merge clusters. Erase one of the modes corresponding to the clusters and append points belonging to j to cluser of i
+					for (int k=0; k < clusters[j].points().size(); k++) {
```
Fluci: The [STL is your friend](http://www.cplusplus.com/reference/vector/vector/insert/): `clusters[i].points().insert(clusters[i].points().begin(), clusters[j].points().begin(), clusters[j].points().end())`.

More efficient, shorter, cleaner.

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
-
-	return clustering;
+
+	return clusters;
+}
+
+double MeanShift::apply_gaussian_kernel(const Eigen::VectorXd point, const Eigen::VectorXd mean) const {
+
+	// check if dimensions match
+	if (mean.size() != point.size()) BOOST_LOG_TRIVIAL(error) << "Vector dimensions don't match";
```
Fluci: Since this method is private, is this check necessary for production code? It should never be called by anything else then by you, which implies it's mostly for debugging. Maybe better use an `assert()`? (asserts are optimized away for release code). Note: If the method was protected or public, you might want to add some error [handling](https://isocpp.org/wiki/faq/exceptions).

Why do I care about the removal of this self-check and not others? It is called many times, so efficiency matters here a lot more than at other places.

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
+	}
+
+	// Initialize some stuff used in the MeanShift loop
+	std::vector<Eigen::VectorXd> prev = points;
+	std::vector<Eigen::VectorXd> curr(points.size());
+	bool has_converged = false;
+  int iterations = 0; //for logging
+
+	// The MeanShift loop
+	while (!has_converged) {
```
Fluci: Let me paraphrase what I read, so we make sure, I understand the code (assume the data will converge): 

- You move each cluster according to the gaussian weighted cog relative to the other clusters.
- You merge clusters, that have a distance bellow a certain threshold.
- You sum up the movement between cluster centers to detect convergence.

Problems I have:
- I'm very afraid of the `n^2` loop inside the while loop, we will have a lot of points so this might take very long.
- You call `iterate_mode(prev[i], prev)`, I've understood mean shift as `iterate_mode(prev[i], points)`. This means: You converge relative to the cluster centers, I expected it to converge according to the original points. I don't know the implications of this. Is this a design decision or a bug? If you contract according to the cluster centers, there's no concept of "weight", so a a cluster consisting of a thousand points will have the same effect as one only with one point. I think this might be undesirable. 
- If you choose to converge with the initial points fixed, you might want to have two phases: 1. for each point, converge independently according to `iterate_mode(.., points);`. 2. Merge all converged points into clusters according to the threshold.
- Convergence detection via sum: I tend to distrust this solution: 
  case A: You have a lot of clusters that move very slightly.
  case B: You have very little clusters that move a lot.
  I claim case A and B can be hard to distinguish when deciding via a sum, since the threshold doesn't scale accordingly. Proposed solution: decide for every cluster center separately if it converged. The overall solution converged if all cluster centers converged.


I'm available via Skype nearly all day, if you want to discuss it in more detail.

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
+			for (int j=i+1; j<curr.size(); j++) {
+				Eigen::VectorXd diff = curr[i] - curr[j];
+
+				if (diff.norm() < MERGE_THRESHOLD) {
+					// Merge clusters. Erase one of the modes corresponding to the clusters and append points belonging to j to cluser of i
+					for (int k=0; k < clusters[j].points().size(); k++) {
+						clusters[i].points().push_back(clusters[j].points()[k]);
+					}
+					//BOOST_LOG_TRIVIAL(debug) << "Merging cluster " << j << " into cluster " << i << ", distance " << diff.norm();
+					clusters.erase(clusters.begin() + j);
```
Fluci: Same bug as in the mock code: if you remove the j-th element, `erase` moves all other elements by one to fill the gap. You need to decrement `j` by one to account for that.

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
+
+	// Initialize Clusters. Initially, every point has its own cluster.
+	std::vector<Cluster> clusters(cloud.size());
+	for (PointIndex i=0; i<cloud.size(); i++) {
+		 clusters[i].points().push_back(i);
+
+	}
+
+	// Initialize some stuff used in the MeanShift loop
+	std::vector<Eigen::VectorXd> prev = points;
```
Fluci: Took me a while to realize, those are cluster centers and not just points. Maybe call them `prevCenters` and `currCenters` or so?

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
-	std::vector<Cluster> clustering = std::vector<Cluster>(_k);
+					break;
+				}
+			}
+		}
 
-  //Initialize RNG
-	srand(time(NULL));
+		if (clusters.size() != curr.size())
+			BOOST_LOG_TRIVIAL(warning) << "Cluster count does not equal mode count. (" << clusters.size() << " != " << curr.size() << ")";
```
Fluci: You should add some remark that this is not an issue the user can fix, but a bug in the code was detected. I usually put something like "Bug: " or "Found internal inconsistency: " in front of it.

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
+	if (mean.size() != point.size()) BOOST_LOG_TRIVIAL(error) << "Vector dimensions don't match";
+
+	// Euclidean distance squared btw. mean of kernel and point of interest
+	const double d = (mean-point).squaredNorm();
+
+	// Apply gaussian distribution
+	return exp(-d/(2*_window_size));
+}
+
+Eigen::VectorXd MeanShift::iterate_mode(const Eigen::VectorXd mode, const std::vector<Eigen::VectorXd>& state) const {
```
Fluci: `state` seems a bit un-sharp to me, maybe `referencePoints` or `fixedPoints`?

lib/clustering/mean_shift.cpp
```diff
@@ -6,39 +6,113 @@
...
-    }
-  }
-	clustering.erase(clustering.begin() + nextInsert, clustering.end());
-
-	return clustering;
+
+	return clusters;
+}
+
+double MeanShift::apply_gaussian_kernel(const Eigen::VectorXd point, const Eigen::VectorXd mean) const {
```
Fluci: `apply_..` seems a bit general, `gaussian_weight()`?

lib/clustering/mean_shift.h
```diff
@@ -9,19 +9,39 @@
...
   /// k is the number of desired clusters
-  MeanShift(const int k);
+  MeanShift(double window_size);
   /// Splits a point cloud into k clusters randomly
   std::vector<Cluster> operator()(const PointCloud& cloud) const;
 
 private:
-    const int _k;
+  // when two peaks are closer than this, they are merged.
+  const double MERGE_THRESHOLD = 0.0001;
```
Fluci: I expect, we want to tweak those parameters interactively in the future. I would prefer public accessors for them.

lib/clustering/mean_shift.h
```diff
@@ -9,19 +9,39 @@
...
+  const double CONVERGENCE_THRESHOLD = 0.1;
+
+  // If algorithm hasn't converged after this amount of iterations, we abort.
+  const int MAX_ITERATIONS = 100;
+
+  // window size parameter for mean shift algorithm
+  double _window_size;
+
+  // contains the means of all the mean shift clusters
+  PointCloud means;
```
Fluci: dead code? since the method is "const", this should probably not exist.

lib/clustering/mean_shift.h
```diff
@@ -9,19 +9,39 @@
...
 
   /// k is the number of desired clusters
-  MeanShift(const int k);
+  MeanShift(double window_size);
   /// Splits a point cloud into k clusters randomly
   std::vector<Cluster> operator()(const PointCloud& cloud) const;
 
 private:
-    const int _k;
+  // when two peaks are closer than this, they are merged.
```
Fluci: tripple slashes "///" if doxygen should show documentation

lib/clustering/mean_shift.h
```diff
@@ -9,19 +9,39 @@
 #include "clustering.h"
 #include <vector>
 #include "generic/cluster.h"
+#include <Eigen/Core>
 
 namespace MouseTrack {
```
Fluci: A short description of the algorithm would be nice, especially since you keep referencing "mean-shift iteration" at other places, but never define the term.

lib/clustering/mean_shift.test.cc
```diff
@@ -1,44 +1,107 @@
 /// \file
```
Fluci, line 1: Interesting, I never realized, how hard it is to verify mean shift via unit tests. It's so easy visually. I think we leave it like that for the moment.

Fluci (commented):
We probably should discuss the core algorithm in a more interactive way (I'm available via Skype until evening).

Logging is mostly fine. I've noted my thoughts on control-flow and error logs at the corresponding places.

Fluci:
One other thing: we still need to think about normalization of dimensions at some point

lib/clustering/mean_shift.cpp
luhug: Right, I used it for debugging. Will remove.

lib/clustering/mean_shift.cpp
luhug: OK, I'll have a look in the evening

lib/clustering/mean_shift.cpp
luhug: An assert is a good idea.

lib/clustering/mean_shift.cpp
luhug: Let's discuss in Skype, I'll let you know when I'm home

lib/clustering/mean_shift.h
luhug: will do

lib/clustering/mean_shift.h
luhug: You're right, forgot to delete. My bad.

lib/clustering/mean_shift.h
luhug: Do we want to show documentation for private methods in the Doxygen as well?

lib/clustering/mean_shift.cpp
luhug: Let's discuss in Skype

lib/clustering/mean_shift.cpp
luhug: I thought about this already. Right after the erase() there is a break; statement, so it should be fine as it is.

lib/clustering/mean_shift.h
Fluci: oh, so that was consciously, I thought you forgot it. I don't mind hiding it. I usually make doc comments available to doxygen (I think it has its own flags where it can hide private stuff if desired).

lib/clustering/mean_shift.cpp
luhug: They could indeed be named better. There is some ambiguity in the naming anyway, sometimes they're also called modes so I might go over the variable naming again.

lib/clustering/mean_shift.cpp
luhug: Will do

lib/clustering/mean_shift.cpp
luhug: Hmm, I agree that the naming is not optimal. But they are not really points in the sense of PointCloud-points, they are just the other modes (cluster centers). Maybe other_modes? ... Haven't found a name that I'm happy with yet.

lib/clustering/mean_shift.cpp
luhug: I'm not sure if we can get past the O(n^2), as the Mean Shift algorithm is stated to be O(n^2) in my references.. If the complexity becomes an issue, i think we'd need to focus on reducing the number of points in the input point cloud in a processing step (e.g. condense close-together points in a single point) and hope this doesn't mess with the mean shift algorithm..

luhug: - On the second point: That is actually a very good point. I have to look at my references again, I might have misinterpreted some variables. Your way acutally makes more sense to me. However it will increase complexity a bit.

- On the third point: I think this will only increase complexity. If two modes are identical (distance < merge threshold) and you don't merge them, you perform the same operations twice in iterate_mode. What would be the advantage with two phases?

- On the fourth point: maybe a simpler solution to the problem would be to use `CONVERGENCE_THRESHOLD*clusters.size()`? This would mean that the average movement is below a certain threshold. Still less accurate than your solution but I think more efficient.

lib/clustering/mean_shift.cpp
Fluci: Ups, didn't notice the break. 

But then there might be a problem with `i`:

Assume centers a,b,c. All closer than MERGE_THRESHOLD to each other.

we then have:
curr = [a,b,c]
i=0, j=1
a is compared to b and they are merged.
curr = [a,c]
break;
next iteration: i = 1, j = 2 (j loop not entered)
i = 2, (i loop not entered)
termination

a was never compared to c

Maybe I did a mistake, executing algorithms by hand is always tricky ...

luhug:
I think there's two ways to do it.
1. Normalize in the registration step specific to our purposes, and use the mean shift algorithm as is.

2. Normalize in the registration step in a general way (e.g. all cordinates btw. 0 and 1) and use a non-radial Gaussian kernel in the mean-shift (In other words, window size is a 4D vector, use different window sizes for the dimensions)

I think 1) is easier but 2) is cleaner.

lib/clustering/mean_shift.cpp
Fluci: postponing to skype...

Fluci:
At the moment, I'd prefer to keep any normalization within the clustering algorithm, since we normalize specifically for the clustering algorithm. This might change, if other modules also need normalized coordinates. Anyway, I propose to focus on the algorithm for the moment and think about normalization at a later point.

lib/clustering/mean_shift.cpp
luhug: This is a valid point - I was only considering two clusters being merged. It's not tragic, as clusters a and c are merged in the next mean shift iteration, but it's definitely more efficient if we change it. I'll have to look at the merge section again anyway.

lib/clustering/mean_shift.h
luhug: For me it doesn't matter, but I think we should be consistent. Probably the easiest way is to show everything and then hide the stuff we don't want afterwards.

itko:
I will start reviewing once you guys have sorted out the ongoing discussions. It's hard to review when there's so many things in the air. @luhug let me know when you think I should start.

lib/clustering/mean_shift.cpp
luhug: What kind exception should I throw here? Is there a standard one for this case or should I just throw a string message?

luhug:
@Fluci @itko I made all the discussed changes from tuesday. Now ready for review.

lib/clustering/mean_shift.cpp
```diff
@@ -32,67 +32,54 @@ std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
...
-					curr.erase(curr.begin() + j);
-
-					break;
-				}
+	Eigen::VectorXd prevCenter;
+	std::vector<Eigen::VectorXd> currCenters = points;
+
+	// For each point...
+	for(PointIndex i = 0; i < clusters.size(); i++) {
+		prevCenter = points[i];
```
Fluci: looks like dead code, it's set again at line 47 but never read. (Probably you started with a while loop?)

lib/clustering/mean_shift.cpp
```diff
@@ -32,67 +32,54 @@ std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
...
-				total_movement += (curr[i]-prev[i]).norm();
+	// Merge step
+	// For every mode..
+	for (int i=0; i<currCenters.size(); i++) {
+		// Check modes we haven't already executed the (i)-loop for
+		for (int j=i+1; j < currCenters.size(); j++) {
+			Eigen::VectorXd diff = currCenters[i] - currCenters[j];
+			if (diff.norm() < _merge_threshold) {
+				// Merge clusters. Erase one of the modes corresponding to the clusters and append points belonging to j to cluser of i
+				clusters[i].points().insert(clusters[i].points().begin(), clusters[j].points().begin(), clusters[j].points().end());
```
Fluci: Note: with the 2 phase approach, this line can be further simplified to `clusters[i].points().push_back(j);` (we always merge j into i, while j > i, hence j is always a cluster as initialized at the start).
There are probably further clean ups, but I'd push them to our optimization phase.

Fluci (approved):

lib/clustering/mean_shift.h
```diff
@@ -9,19 +9,55 @@
...
 private:
-    const int _k;
+  /// when two peaks are closer than this, they are merged. Must be larger than _convergence_threshold for convergence.
+  double _merge_threshold = 0.001;
+
+  /// Convergence is decided if modes haven't moved more than this on average. Must be smaller than _merge_threshold for convergence.
+  double _convergence_threshold = 0.0001;
+
+  /// If algorithm hasn't converged after this amount of iterations, we abort.
+  int _max_iterations = 100;
```
itko, line 51: Perhaps there should be a warning or something like: "Number of iterations not set. Defaults to 100.". Not sure if necessary.

itko (approved):

lib/clustering/mean_shift.h
luhug: Postponing to optimisation phase if it's ok.

PR #15, opened 2018-04-04
Getting the baseline descripting and matching

itko:
1) COG descripting
2) Matching using nearest neighbours

Still no unit tests

### DIFF
--- a/lib/descripting/cog.cpp
+++ b/lib/descripting/cog.cpp
@@ -10,10 +10,12 @@
 
 namespace MouseTrack {
 
+CenterOfGravity::~CenterOfGravity(){};
+
 /// Takes a cluster and its referencing PointCloud and creates a cluster
 /// descriptor
-std::unique_ptr<ClusterDescriptor> CenterOfGravity::operator()(
-    const Cluster &cluster, const PointCloud &cloud) const {
+std::unique_ptr<ClusterDescriptor> CenterOfGravity::
+operator()(const Cluster &cluster, const PointCloud &cloud) const {
   Coordinate sumX = 0, sumY = 0, sumZ = 0;
   size_t numPoints = cluster.points().size();
   // Go through each point in the cluster
@@ -29,4 +31,4 @@ std::unique_ptr<ClusterDescriptor> CenterOfGravity::operator()(
   return descriptor;
 };
 
-}  // namespace MouseTrack
+} // namespace MouseTrack
--- a/lib/descripting/cog.cpp
+++ b/lib/descripting/cog.cpp
@@ -10,19 +10,20 @@
 
 namespace MouseTrack {
 
-/// Takes a cluster and its referencing PointCloud and creates a cluster descriptor
-std::unique_ptr<ClusterDescriptor> CenterOfGravity::operator()(const Cluster& cluster, const PointCloud& cloud) const {
-	Coordinate sumX = 0, sumY = 0, sumZ = 0;
-	size_t numPoints = cluster.points().size();
-	for (PointIndex i = 0; i < cluster.points().size(); i++) {
-		 sumX += cloud[i].x();
-		 sumY += cloud[i].y();
-		 sumZ += cloud[i].z();
-	}
-	std::unique_ptr<ClusterDescriptor> descriptor(new Center(sumX/numPoints, sumY/numPoints, sumZ/numPoints));
-	return descriptor;
+/// Takes a cluster and its referencing PointCloud and creates a cluster
+/// descriptor
+std::unique_ptr<ClusterDescriptor> CenterOfGravity::
+operator()(const Cluster &cluster, const PointCloud &cloud) const {
+  Coordinate sumX = 0, sumY = 0, sumZ = 0;
+  size_t numPoints = cluster.points().size();
+  for (PointIndex i = 0; i < cluster.points().size(); i++) {
+    sumX += cloud[i].x();
+    sumY += cloud[i].y();
+    sumZ += cloud[i].z();
+  }
+  std::unique_ptr<ClusterDescriptor> descriptor(
+      new Center(sumX / numPoints, sumY / numPoints, sumZ / numPoints));
+  return descriptor;
 };
 
-} // MouseTrack
-
-
+} // namespace MouseTrack
--- a/lib/descripting/cog.cpp
+++ b/lib/descripting/cog.cpp
@@ -17,7 +17,7 @@ std::unique_ptr<ClusterDescriptor> CenterOfGravity::operator()(
   Coordinate sumX = 0, sumY = 0, sumZ = 0;
   size_t numPoints = cluster.points().size();
   // Go through each point in the cluster
-  for (PointIndex i = 0; i < cluster.points().size(); i++) {
+  for (PointIndex i : cluster.points()) {
     // Get the sum of all coordinates along x, y ,z
     sumX += cloud[i].x();
     sumY += cloud[i].y();
--- a/lib/descripting/cog.cpp
+++ b/lib/descripting/cog.cpp
@@ -12,18 +12,21 @@ namespace MouseTrack {
 
 /// Takes a cluster and its referencing PointCloud and creates a cluster
 /// descriptor
-std::unique_ptr<ClusterDescriptor> CenterOfGravity::
-operator()(const Cluster &cluster, const PointCloud &cloud) const {
+std::unique_ptr<ClusterDescriptor> CenterOfGravity::operator()(
+    const Cluster &cluster, const PointCloud &cloud) const {
   Coordinate sumX = 0, sumY = 0, sumZ = 0;
   size_t numPoints = cluster.points().size();
+  // Go through each point in the cluster
   for (PointIndex i = 0; i < cluster.points().size(); i++) {
+    // Get the sum of all coordinates along x, y ,z
     sumX += cloud[i].x();
     sumY += cloud[i].y();
     sumZ += cloud[i].z();
   }
-  std::unique_ptr<ClusterDescriptor> descriptor(
+  // Divide to get average x, y, z. This is the center of gravity.
+  std::unique_ptr<Center> descriptor(
       new Center(sumX / numPoints, sumY / numPoints, sumZ / numPoints));
   return descriptor;
 };
 
-} // namespace MouseTrack
+}  // namespace MouseTrack
--- a/lib/descripting/cog.cpp
+++ b/lib/descripting/cog.cpp
@@ -12,7 +12,14 @@ namespace MouseTrack {
 
 /// Takes a cluster and its referencing PointCloud and creates a cluster descriptor
 std::unique_ptr<ClusterDescriptor> CenterOfGravity::operator()(const Cluster& cluster, const PointCloud& cloud) const {
-	std::unique_ptr<ClusterDescriptor> descriptor(new Center());
+	Coordinate sumX = 0, sumY = 0, sumZ = 0;
+	size_t numPoints = cluster.points().size();
+	for (PointIndex i = 0; i < cluster.points().size(); i++) {
+		 sumX += cloud[i].x();
+		 sumY += cloud[i].y();
+		 sumZ += cloud[i].z();
+	}
+	std::unique_ptr<ClusterDescriptor> descriptor(new Center(sumX/numPoints, sumY/numPoints, sumZ/numPoints));
 	return descriptor;
 };
 
--- a/lib/generic/center.cpp
+++ b/lib/generic/center.cpp
@@ -10,27 +10,20 @@
 
 namespace MouseTrack {
 
-const Coordinate& Center::x() const {
-	return this->_x;
-}
+const Coordinate &Center::x() const { return this->_x; }
 
-const Coordinate& Center::y() const {
-	return this->_y;
-}
+const Coordinate &Center::y() const { return this->_y; }
 
-const Coordinate& Center::z() const {
-	return this->_z;
-}
+const Coordinate &Center::z() const { return this->_z; }
 
-double Center::compare(const ClusterDescriptor* other) const {
-	const Center * otherCenter = dynamic_cast<const Center*>(other);
-	// Get Euclidean distance
-	double err = pow((this->_x - otherCenter->x()),2) + pow((this->_y - otherCenter->y()),2) + pow((this->_z - otherCenter->z()),2);
-	double n = sqrt(err);
-	return n;
+double Center::compare(const ClusterDescriptor *other) const {
+  const Center *otherCenter = dynamic_cast<const Center *>(other);
+  // Get Euclidean distance
+  double err = pow((this->_x - otherCenter->x()), 2) +
+               pow((this->_y - otherCenter->y()), 2) +
+               pow((this->_z - otherCenter->z()), 2);
+  double n = sqrt(err);
+  return n;
 };
 
-
-} // MouseTrack
-
-
+} // namespace MouseTrack
--- a/lib/generic/center.cpp
+++ b/lib/generic/center.cpp
@@ -10,6 +10,10 @@
 
 namespace MouseTrack {
 
+Center::Center(Coordinate x, Coordinate y, Coordinate z)
+    : _x(x), _y(y), _z(z) {}
+Center::~Center() {}
+
 const Coordinate &Center::x() const { return this->_x; }
 
 const Coordinate &Center::y() const { return this->_y; }
--- a/lib/generic/center.cpp
+++ b/lib/generic/center.cpp
@@ -26,4 +26,4 @@ double Center::compare(const ClusterDescriptor *other) const {
   return n;
 };
 
-} // namespace MouseTrack
+}  // namespace MouseTrack
--- a/lib/generic/center.cpp
+++ b/lib/generic/center.cpp
@@ -12,6 +12,7 @@ namespace MouseTrack {
 
 Center::Center(Coordinate x, Coordinate y, Coordinate z)
     : _x(x), _y(y), _z(z) {}
+
 Center::~Center() {}
 
 const Coordinate &Center::x() const { return this->_x; }
@@ -30,4 +31,4 @@ double Center::compare(const ClusterDescriptor *other) const {
   return n;
 };
 
-}  // namespace MouseTrack
+} // namespace MouseTrack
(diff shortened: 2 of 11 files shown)


itko:
@Fluci @luhug This still does not compile due to a mistake in the Nearest Neighbours. Can you explain what I am doing wrong?

Fluci:
Found it, that's an annoyance of `std::map`. The bracket operator `[]` of map is not available as const, according to the [documentation](http://www.cplusplus.com/reference/map/map/operator[]/) it creates a new entry in the map, should a key not be found. In my opinion a questionable design decision (std vector does provide a const version of []). You need to use ".find(lastFrameNumber)->second" (make sure lastFrameNumber exists, otherwise second will hold random/empty/undefined data). Now it should compile.

lib/descripting/cog.cpp
```diff
@@ -12,7 +12,14 @@ namespace MouseTrack {
 
 /// Takes a cluster and its referencing PointCloud and creates a cluster descriptor
 std::unique_ptr<ClusterDescriptor> CenterOfGravity::operator()(const Cluster& cluster, const PointCloud& cloud) const {
-	std::unique_ptr<ClusterDescriptor> descriptor(new Center());
+	Coordinate sumX = 0, sumY = 0, sumZ = 0;
+	size_t numPoints = cluster.points().size();
+	for (PointIndex i = 0; i < cluster.points().size(); i++) {
+		 sumX += cloud[i].x();
```
Fluci: Probably this?
```C++
for(PointIndex i : cluster.points())
```

lib/descripting/cog.cpp
```diff
@@ -10,12 +10,23 @@
...
-	std::unique_ptr<ClusterDescriptor> descriptor(new Center());
-	return descriptor;
+/// Takes a cluster and its referencing PointCloud and creates a cluster
+/// descriptor
+std::unique_ptr<ClusterDescriptor> CenterOfGravity::operator()(
+    const Cluster &cluster, const PointCloud &cloud) const {
+  Coordinate sumX = 0, sumY = 0, sumZ = 0;
+  size_t numPoints = cluster.points().size();
+  // Go through each point in the cluster
+  for (PointIndex i = 0; i < cluster.points().size(); i++) {
```
Fluci: Probably `for(PointIndex i : cluster.points())`,

lib/descripting/cog.cpp
```diff
@@ -10,12 +10,23 @@
...
+    // Get the sum of all coordinates along x, y ,z
+    sumX += cloud[i].x();
+    sumY += cloud[i].y();
+    sumZ += cloud[i].z();
+  }
```
luhug, line 27: For optimization phase: we can use Eigen::Vector3d here.

lib/generic/center.cpp
```diff
@@ -6,14 +6,24 @@
...
-} // MouseTrack
+const Coordinate &Center::z() const { return this->_z; }
 
+double Center::compare(const ClusterDescriptor *other) const {
+  const Center *otherCenter = dynamic_cast<const Center *>(other);
+  // Get Euclidean distance
+  double err = pow((this->_x - otherCenter->x()), 2) +
+               pow((this->_y - otherCenter->y()), 2) +
+               pow((this->_z - otherCenter->z()), 2);
+  double n = sqrt(err);
```
luhug, line 30: For optimization phase: use Eigen::norm()

lib/generic/center.h
```diff
@@ -8,20 +8,29 @@
...
 public:
-	/// Returns a distance/norm `n` between `this` and `other`.
-	/// It should be small for two instances that are "similar" to each other,
-	/// and large for two instances that are very different.
-	/// Important properties:
-	/// 1. 0 <= n <= inf
-	/// 2. if this == other -> n == 0
-	virtual double compare(const ClusterDescriptor* other) const;
-};
+  Center(Coordinate x, Coordinate y, Coordinate z) : _x(0), _y(0), _z(0) {}
```
Fluci: You probably wanted: `_x(x), _y(y), _z(z)`

Second: I know it's nearly nothing, but I also would push the Constructor and Destructor implementation to the .cpp file and only leave the declaration here.

lib/generic/cluster_descriptor.h
```diff
@@ -7,18 +7,18 @@
...
 class ClusterDescriptor {
 public:
-	/// Returns a distance/norm `n` between `this` and `other`.
-	/// It should be small for two instances that are "similar" to each other, 
-	/// and large for two instances that are very different.
-	/// Important properties:
-	/// 1. 0 <= n <= inf
-	/// 2. if this == other -> n == 0
-	virtual double compare(const ClusterDescriptor* other) const = 0;
+  virtual ~ClusterDescriptor(){};
```
Fluci: Thanks for adding :+1: 
Can you move the implementation to the cpp? Moving as much as possible into the cpp saves compile time.

lib/matching/matching.h
```diff
@@ -5,23 +5,25 @@
...
 
 class Matching {
 public:
-    /// Given a list of cluster chains, we find the best matching to append the given list of descriptors.
-    /// descriptors: a list of descriptors from a new frame
-    /// chains: a list of cluster chains to which we want to appeand one cluster from descriptors
-    /// returns a vector of size descriptors.size(), which holds the index of the matching chain for each descriptor.
-    /// If no match was found, the vector value is set to -1.
-    virtual std::vector<long> operator()(const std::vector<std::shared_ptr<const ClusterDescriptor>>& descriptors, const std::vector<ClusterChain>& chains) = 0;
+  ~Matching(){};
```
Fluci: Thanks for adding, implementation in `.cpp`

luhug (approved):

lib/matching/nearest_neighbour.cpp
```diff
@@ -6,13 +6,40 @@
...
+    const std::vector<ClusterChain> &chains) {
+  // Prepare an empty vector that will store the index of each chain
+  std::vector<long> matchIndex;
+  for (size_t j = 0; j < descriptors.size(); j++) {
+    // Initialize a maximum distance
+    double nearest = std::numeric_limits<double>::max();
+    for (size_t i = 0; i < chains.size(); i++) {
+      // In this case we are only really concerned with the last cluster. Get
+      // the frame.
+      FrameNumber lastFrameNumber = chains[i].descriptors().rbegin()->first;
```
Fluci: What if the chain has no descriptors?

1. Either check for this case in the algorithm.
2. Or add a note in the comment of the public interface, that no empty chains are allowed.

I tend to go with 2.: It's more efficient and an easy thing to fulfill.
Arguments for 1: It's more robust and more general, it behaves in an intuitive way.

lib/matching/nearest_neighbour.cpp
```diff
@@ -6,13 +6,40 @@
...
+      // Get the cluster for that frame.
+      std::shared_ptr<const ClusterDescriptor> lastDescriptor =
+          chains[i].descriptors().find(lastFrameNumber)->second;
```
Fluci: `auto descriptor = chains[i].descriptors().rbegin();` and then use `descriptor->first/second`? Improves readability.

lib/matching/matching.h
```diff
@@ -5,23 +5,25 @@
...
+  /// Given a list of cluster chains, we find the best matching to append the
```
Fluci: Automatic reformatting moved line breaks which makes it hard to read, maybe just add two line breaks?

lib/matching/nearest_neighbour.cpp
```diff
@@ -6,13 +6,40 @@
...
+      // Compare the descriptor to the cluster. If it's closer than the previous
+      // chain then choose that one.
+      // TODO This means that there may be repeated descriptors pointing to the
```
Fluci: True, maybe go the other way around? For each chain, look for the closest descriptor, all other descriptors get -1 entries (a  new chain will be generated for those entries by the pipeline).

lib/matching/nearest_neighbour.h
```diff
@@ -11,14 +11,20 @@
...
-		/// returns a vector of size descriptors.size(), which holds the index of the matching chain for each descriptor.
-		/// If no match was found, the vector value is set to -1.
-		virtual std::vector<long> operator()(const std::vector<std::shared_ptr<const ClusterDescriptor>>& descriptors, const std::vector<ClusterChain>& chains);
+class NearestNeighbour : public Matching {
+
+public:
+  virtual ~NearestNeighbour() {}
+
+  /// Given a list of cluster chains, we find the best matching to append the
+  /// given list of descriptors. descriptors: a list of descriptors from a new
```
Fluci: Automatic formatting messed with newlines, I think, I need to place two line-breaks in the future.

lib/descripting/cog.cpp
Fluci: I would definitely not change it at the moment. Clang (and other compilers) might be able to vectorize this loop. In that case it might be more efficient than creating Eigen vectors.

lib/generic/center.h
itko: What's the benefit? Just style?

lib/generic/cluster_descriptor.h
itko: You would really prefer adding a whole .cpp file just to implement an empty destructor?

lib/matching/matching.h
itko: Same as above. There's no .cpp currently

lib/matching/nearest_neighbour.cpp
itko: What is the possibility here: that the descriptor map is empty or that the descriptor pointer in the map is invalid/null?

lib/matching/nearest_neighbour.cpp
itko: I like it!

lib/matching/nearest_neighbour.cpp
itko: The consequence would be that instead of the same cluster chain having multiple matches we have the same descriptor having multiple matches. Both are not ideal but I agree, the latter is much preferred.

lib/matching/nearest_neighbour.h
itko: What do you mean? I like it this way. Either way, the LLVM styler automatically cuts two lines to one line.

itko:
Still didn't move some Constructors/Destructors to the .cpp. It seems a bit awkward to move them there when they are empty AND they are the only functions that are being implemented. But I can do that without a problem, just want to hear your feedback.

Fluci:
Maybe I have to clarify the problem:

Each cpp file compiles independently. That's why you can use `make -j`.
`void f()` is a declaration. It tells the compiler: "hey, look out, somewhere is a function `f`, don't panic if you don't know where, we'll fix that later", the compiler puts a mental note somewhere and goes ahead without generating binary code.
`void f(){...}` is a definition. It tells the compiler: "Look, here is `f`, it should do ...", the compiler parses  and generates binary code for its body.

What happens if you have multiple definitions of `f`, i.e. multiple places in your binary that correspond to the binary code of `f`? I'm not sure (please check with the standard to be sure). I thought, it was ok, as long as they do the exact same thing. But: since we compile cpp's independently during development, this assumption might be violated and we end up with hard to track bugs. 

Header files are just copy+pasted into source files. If you put definitions into header files, those definitions are parsed and created each time a source file uses a header file. So we end up with redundant parts in the binary that could create unnecessary problems.

In short: 

Con moving it:
- It's hardly any code.

Pro moving it: 
- Compiler has to less work.
- Compiler creates no redundant code in the final binary.
- Weird bugs are avoided.
- When we need to add something to the method, we have to move it anyway.

Exeption:

You can't put templated stuff into source files: `template <typename A> void f(A a){...}` is a definition but belongs into the header files. The reason is, that templates can't be generated without knowing their types in advance. Hence they have to be in the header file and the compiler takes care of everything.

One more beautiful weirdness of C++ :)

lib/matching/nearest_neighbour.h
Fluci: See line 21: "frame chains:" frame belongs to the previous line, it was merged with the next line which makes it really hard to read. Same thing line 20: "... descriptors. descriptors: " and line 22 "cluster from descriptors returns a vector"

I have no problem, breaking lines to 80 character width.  It's more about merging lines in such a way, they get hard to read.

itko:
@Fluci Good points you made. Check out my fixes. I hope I didn't miss anything.

lib/matching/matching.h
```diff
@@ -14,13 +14,17 @@ namespace MouseTrack {
...
+  virtual ~Matching();
+
   /// Given a list of cluster chains, we find the best matching to append the
   /// given list of descriptors. descriptors: a list of descriptors from a new
   /// frame chains: a list of cluster chains to which we want to appeand one
   /// cluster from descriptors returns a vector of size descriptors.size(),
   /// which holds the index of the matching chain for each descriptor. If no
   /// match was found, the vector value is set to -1.
+  ///
+  /// Note: each cluster chain must have a descriptor for each frame of the
```
Fluci: Nope: each cluster chain must have at least one descriptor and cluster pair belonging to the same frame.

Cluster chains are allowed to have holes. Forbidding holes, would be far to brittle. Just think about all the noise in the data. The clustering could create 10 clusters for frame `i` but 9 cluster for frame `i+1`.

(also see further below the more detailed considerations)

lib/matching/nearest_neighbour.cpp
```diff
@@ -11,29 +11,32 @@
...
+    // Get the frame number and the descriptor
+    auto descriptor = chains[i].descriptors().rbegin();
+    // Get the descriptor for that frame.
+    auto lastDescriptor = descriptor->second;
+    for (size_t j = 0; j < descriptors.size(); j++) {
       // Compare the descriptor to the cluster. If it's closer than the previous
       // chain then choose that one.
-      // TODO This means that there may be repeated descriptors pointing to the
-      // same cluster chain.
+      // TODO This means that there may be multipel chains pointing to the
```
Fluci: Hmm, we really need to think about the exact semantics we want for this.

1. If we allow cluster chains to share the same descriptor, this means they will be equal from that point on. So no point in having two cluster chains (assuming we only consider the last descriptor). Assuming clusters correspond in some way to real world objects, what would it mean if they merge? Weird ...

2. If we allow descriptors to share the same cluster chain: Same as at 1, what does it mean for a real world object to split?

3. If we allow for holes in the chains and require each cluster to be assigned to exactly one chain, we have options:
  a. If we have a 1:1 match, we could decide if it is actually close enough (thresholding or other fancy stuff).
  b. If we have multiple descriptors for one chain, we could choose the closest one.
  c. If we have multiple cluster chains for one descriptor, we could also assign it to the closest chain.
  d. We can take the frame number into account, and skip very old chains (chains that weren't modified during the last k frames). We might need to pass in the active frame index in the method at some point.
  e. We can try a lot other heuristics up to bipartite graph matching.

I'm strongly in favor for 3. It captures the semantics very well: noise could create cluster chains and we are able to "forget" them. 

The question is, how to fix nearest neighbor if we want 3, hmm ...

Maybe do two runs? The first run collects sets of chains for each descriptor: each chain `c` says "I'm closest to descriptor `a`", so `c` is added to the set of descriptor `a`.
In a second run, each descriptor chooses the closest chain from its set.
One could analyze, if there are still unassigned descriptors and chains and iterate, but that seems overkill to me at the moment.

We will clearly loose stuff, that we don't have to. But since the entire matching stuff wanders at the edges of NP-completeness anyway, It should be fine for the moment.


Your thoughts?

Fluci:
We first need to pin the semantics down, see my comment for this.

I took the freedom, to reformat the doc comments.

lib/matching/matching.h
itko: That's why I said each frame of the chain, not each frame of the stream! Perhaps a better term would be each "valid frame" of the chain? Each "filled frame"?  "Active frame"?

lib/matching/nearest_neighbour.cpp
itko: > Maybe do two runs? The first run collects sets of chains for each descriptor: each chain c says "I'm closest to descriptor a", so c is added to the set of descriptor a.
> In a second run, each descriptor chooses the closest chain from its set.

Given our time constraints for the baseline, this is the best balance between implementation time and results. But, like you said, it basically guarantees we will lose some information.

Another method that could be implemented in the future accounting for some historical data. At 30fps I would say that descriptors shouldn't change too much between frames. I envision some sort of moving average of distance between descriptor and chain cluster. Higher weight given to more recent frames.

Let me know your thoughts. As mentioned, I think the two-run approach is most feasible at the moment. But this is a prime candidate for improvements in the next phase.

lib/matching/nearest_neighbour.h
itko: I understand now. Fixed this case, will try to fix them as I see them

lib/matching/matching.h
Fluci: Oh, ok, then it's ok, if "each frame of the chain" allows holes in the sequence. Thanks for clarifying.

lib/matching/nearest_neighbour.cpp
Fluci: A moving average on constant cluster attributes sounds very promising, i like it.

For the moment: can you make sure, descriptors aren't assigned to multiple chains? Can be a very simple solution like a vector of flags and they are only added when no flag is set or so. I just want to avoid merges/splits of chains for the baseline.

I think, we should be fine then.

lib/matching/nearest_neighbour.cpp
```diff
@@ -18,29 +18,31 @@ std::vector<long> NearestNeighbour::operator()(
...
-      long match = 0;
-      double distance = descriptors[j]->compare(lastDescriptor.get());
-      if (nearest > distance) {
-        // Better match than previous chain. Pick that one.
-        match = i;
-        nearest = distance;
+    long match = -1;
+    for (size_t j = 0; j < chains.size(); j++) {
+      // Check if we already assigned this chain
+      if (std::find(matchIndex.begin(), matchIndex.end(), j) ==
```
Fluci: Probably takes O(n). The usage of an additional map `std::map<chainIndex, descriptorIndex>` (or std::unordered_map) could reduce it to O(log n) add the added cost of additional memory usage and a slightly more complicated `(nearest > distance)` update. But ok for the moment.

Fluci (approved):
Can be merged.

PR #16, opened 2018-04-05
First implementation of automatic code style

itko:
This requires some manual work of placing files in the right directories. See the README.md

I recommend setting up your IDE to automatically run clang-format on every file save. This way you're accustomed to the style.

**If you don't like the style, then feel free to change it. The git commit hook explains how to do so.** Let's just make sure we are all using the same style.


setup/README.md
```diff
@@ -0,0 +1,6 @@
+# Instructions
+
+1. The clang-format binary is what styles your code automatically. It needs to placed somewhere in your path. I place it in ~/.local. I am running a Ubuntu system. For other systems maybe it's best to look up online how to install clang-format.
```
Fluci, line 3: There's a package "clang-format" in apt-get. Any reason we should add the executable in our repository and not just refer to it? Shipping an executable implies a certain environment to run (not very universal) and it requires their License file.

Fluci (commented):
How do I know, it did work? I've tried to modify `main.cpp` and commit, but it looked exactly the same as before.

Fluci:
I don't know the difference between different styles, I would like to go with your recommendation. What do you recommend? Does it use LLVM by default?

How to proceed?

Is it worth to create a "reformat" commit, once we have a style and it's running for everybody? I would like to avoid huge diffs on reviews where I just had to fix some lines in an old file.

setup/README.md
itko: I haven't tried it by installing. I think I initially tried to install 3.4, which is no longer in the apt cache, so I figured it just doesn't exist as a package in Ubuntu 16.04. . I didn't realize there were newer versions in the cache until later. Can you try installing and seeing if it works? I will update accordingly.

itko:
1) LLVM is good enough for me for now. See here for a very nice comparison of all of them: [https://github.com/motine/cppstylelineup](https://github.com/motine/cppstylelineup)

Google style is also nice. Mozilla is awful imo. Ideally we should use the style that is most good to get used to, i.e. the one that is used in most industries. We can always change later. The key is that we all have the same style. 

2) Yes. I was hoping to have one big commit that changes the style of all files (running clang-format on all files). But to avoid annoying merge conflicts, I thought the best time would be during the Assessment phase of our project. I.e. in a few days

itko:
You initially commented that the git hook did not seem to work for you. Is that still the case? When you commit, you should see in the terminal a list of the files that clang-format edited

Fluci:
1. oh, nice line up. Interesting, LLVM removes the semi-colon at the end of a class declaration, I thought it was necessary with C++. We'll see how it works. How much does clang-format change? does it change the naming of variables? (going from "_age" to "Age"?)

2. Good plan.

3. A list of files, ok, I'll keep an eye on it. By now, I found a plugin for qtCreator and figured out how to run clang-format manually. I'll find a solution.

setup/README.md
Fluci: There are multiple versions (clang-format-3.5 up to clang-format-5.0). I installed the default clang-format. Might be the reason my git hook doesn't work.

itko:
Nope, it does not change naming of variables. I thought we wanted to avoid that since it's just going to slow us down if we dig into naming and stuff like that. We can look into it in the next iteration, if you'd like. Clang-format is basically just white space changes. But it already makes a huge difference.

luhug (approved):

itko:
@Fluci Approve if you think it's ready. Or let me know if there's something to be done. Would be nice if you could update the README.md with any installation issues you had. OR just tell me and I can update.

Fluci:
I still want to get rid of the executable, I just haven't had time yet to try anything with hooks.

itko:
@luhug If it doesn't work for you AND you haven't set it up to automatically run through your IDE, then let me know and I can fix it.

If you are at least able to get it to work on your IDE (for example, that it runs whenever you save the file) then let me know and I can push this.

Fluci:
don't forget to remove the binary, thx

PR #17, opened 2018-04-06
Pipeline implementation added

Fluci:
For time pressure reasons, I've pushed a bit more than just the pipeline:

- Pipeline implementation
- Observer interface for communication with gui
- CommandLineInterface draft for integration testing

### DIFF
--- a/app/pipeline.cpp
+++ b/app/pipeline.cpp
@@ -4,22 +4,325 @@
 ///
 
 #include "pipeline.h"
+#include <boost/log/trivial.hpp>
+#include <iostream>
 
 namespace MouseTrack {
 
-Pipeline::Pipeline() {
+Pipeline::Pipeline() :
+    _controller_should_run(false),
+    _controller_running(false),
+    _controller_terminated(true) {
+    BOOST_LOG_TRIVIAL(debug) << "Constructing pipeline.";
 }
 
-Pipeline::~Pipeline() {
+Pipeline::Pipeline(
+    std::unique_ptr<Reader> reader,
+    std::unique_ptr<Registration> registration,
+    std::unique_ptr<Clustering> clustering,
+    std::unique_ptr<Descripting> descripting,
+    std::unique_ptr<Matching> matching,
+    std::unique_ptr<TrajectoryBuilder> trajectoryBuilder) :
+    _controller_should_run(false),
+    _controller_running(false),
+    _controller_terminated(true),
+    _reader(std::move(reader)),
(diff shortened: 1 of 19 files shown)


app/pipeline.cpp
```diff
@@ -4,22 +4,325 @@
...
+    std::unique_ptr<Reader> reader,
+    std::unique_ptr<Registration> registration,
+    std::unique_ptr<Clustering> clustering,
+    std::unique_ptr<Descripting> descripting,
+    std::unique_ptr<Matching> matching,
+    std::unique_ptr<TrajectoryBuilder> trajectoryBuilder) :
+    _controller_should_run(false),
+    _controller_running(false),
+    _controller_terminated(true),
+    _reader(std::move(reader)),
```
itko, line 29: Just curious: Why is this better than taking a reference to the reader as an argument and initializing via  `_reader(reader)`

app/pipeline.h
```diff
@@ -48,6 +71,70 @@ class Pipeline {
...
+        for(PipelineObserver* o : _observers){
+            lambda(o);
+        }
+    }
+
+    /// Helper for move operations: moves all relevant data members (not mutexes, not threads ...)
+    void moveMembersFrom(Pipeline& p);
+
+    /// Like `start()`, but does not lock thread
+    void _start();
```
itko, line 117: Is this a standard naming convention? Should it not be made more explicit in the name that one locks and the other doesn't?

itko (approved):

app/pipeline.cpp
Fluci: You won't be able to take a reference to a unique_ptr and assign it to a member. std::unique_ptr has no copy-assignment. (That has to do with the semantics of the class: there should always be only one unique_ptr to an object).

The move tells the compiler, that we want to consume the argument. This means, the compiler is allowed to take the member-values from the source (the parameter) and copy them to the target members. The important thing is, that the source members are set to some valid but otherwise undefined values. In our case, we need it for the semantics of std::unique_ptr. In many cases you want to move, because it is more efficient than to copy. For further information, I recommend to google for "Move semantics".

app/pipeline.h
Fluci: I've seen it before while browsing implementations of the STL. In libraries you often have one private method, providing the core functionality, and one or more public methods, providing valid access to them.

luhug (approved):

itko:
@Fluci Merge when ready

PR #18, opened 2018-04-06
Added Trajectory Builder Baseline

luhug:
This essentially just calculates the COG again.. In the optimization phase we can look into centralizing/caching the cog calculation as we use it multiple times


Fluci:
I'll put a post-it in trello for the refactoring, we definitely should fix that redundancy later on.

Fluci (approved):

itko (approved):
I agree that this COG thing needs to be figured out.

I'm also not sure the Trajectory and TrajectoryBuilder flow makes complete sense but I skimmed over it. Let's discuss.

PR #20, opened 2018-04-08
Writer for pipeline

Fluci:
# Writer

A first iteration of a pipeline writer. It writes all data of the in-between steps to a specified output directory. This should simplify debugging. Exception: ClusterDescriptors have no public attributes yet, so they are ignored.

To use it, run `mousetrack` with the additional arguments `-o <outputDir>`.

The writer is added as a PipelineObserver.

I've tried for about an hour to find a suitable read/write 3d file framework, Assimp and openscenegraph looked very promising, but turned out to be far to hard to use for our simple use case. So I wrote our own point cloud exporter in ten minutes. It uses the ascii version of the ply file format.

# CommandLineInterface

I also added a flag for the pipline timer. I think it is more intuitive to parse and set up common values in main, instead to pass the responsibility on to the controllers. I would prefer to push this functionality into the PipelineFactory, but that  requires some refactoring regarding ownership of the observers. An argument against it, would be that the gui controller would create new pipelines anyway each time parameters are changed. It would make sense for it to be responsible for it's own observers. We should discuss this once the gui is stable.

# Bug fix in Registration

Disparity maps are read with values in [0,1]. The original script expected values in [0,255]. I adjusted this assumption.

Edit: I also replaced Eigen's HouseholderQR with a simple inverse. It's not a big deal for release (going from 140 ms to about 80 ms on my machine). But the debug version goes from about 8 seconds to 600 ms. Visual inspection of the point cloud showed no significant change.


app/pipeline_timer.cpp
```diff
@@ -78,7 +83,7 @@ void PipelineTimer::stopTimer(FrameIndex f, const std::string& key){
     auto duration = std::chrono::system_clock::now() - start->second;
     _starts.erase(start);
     int d = std::chrono::duration_cast<std::chrono::microseconds>(duration).count();
-    BOOST_LOG_TRIVIAL(debug) << "(" << f << ", " << key << "): " << d/1000.0 << " milliseconds";
+    BOOST_LOG_TRIVIAL(info) << "(" << f << ", " << key << "): " << d/1000.0 << " milliseconds" << std::flush;
```
Fluci: remove flush

itko (approved):

luhug (approved):

PR #22, opened 2018-04-09
Updated pipeline.cpp to acommodate trajectory builder interface change

luhug:
I changed the trajecotry_builder.h interface when I made the trajectory builder, but pipeline.cpp still used the old one. This is now fixed.


luhug:
Forget it, Felice already fixed it in #19

PR #23, opened 2018-04-09
Implemented Random sampling

luhug:
Implemented Random Sampling as a function to be used in the pipeline before the Mean Shift stage.


lib/generic/random_sample.h
```diff
@@ -0,0 +1,14 @@
...
+///
+///
+
+#pragma once
+
+#include "generic/point_cloud.h"
+
+namespace MouseTrack {
+/// Returns a random sample of size "size" of points from a point cloud
+PointCloud random_sample(PointCloud& cloud, int size);
```
Fluci: missing `const` on cloud

Fluci (approved):
Oh, rejection sampling, should actually be a decent choice since `cloud.size() >> size`. :+1: 

I think, the entry for the cpp in `lib/CMakeLists.txt` is missing. Could you add it and make sure it compiles?

itko (approved):

PR #24, opened 2018-04-11
Feature/fs/acceleration clustered writer

Fluci:
# Uniform grid acceleration structure

Concept: we take a uniform grid (not cartesian, all axes have the same intervals but don't need to be 1).
Each point of a point cloud will fall in a grid cell. Hence we store a list of points for each grid cell.
We can answer a range query (give all points q_i with ||q_i - p|| < r) by looking at the cell for p and all its neighbors closer than r. 

The implementation is not as fast as I would like it, but I've spent far too many hours on it during the last two days (n-dimensional grids are a lot harder than I expected). I would like to push improvements into the optimization phase. We should also look at rougher alternatives like approximate nearest neighbors for really high dimensional spaces.

There's also a BruteForce implementation for the spatial oracle (Interface). It calculates the same queries by simple for-loops. It's a reference implementation to test the UniformGrid against (on TODO list).

## Changes in clustering

I've moved stuff around in clustering, and changed it quite heavily to make it decently efficient for the acceleartion structure. I hope that is ok, @luhug .

# Point Cloud Filtering added

The preprocessing step where we subsample the point cloud as discussed.
The observers for `newFilteredPointCloud`are only called, when a postprocessing step is set. @luhug: We might need to discuss this, depending on what behavior is more suitable for the gui.

# RGB added to Point Cloud

In preparation for cluster writer. It implied an interface change of PointCloud points. I adjusted the according source code locations. (I'm also trying a new style of accessors: const Val& x() for read, Val x(_new) for write (the returned value is undefined at the moment, would it make sense to return the replaced value??).

# Writer for clustered point cloud

As discussed, I've added the feature, that together with a clustering, a point cloud is written who's points have a coloring depending on the cluster. At the moment the points are colored very rudimentary: (c_norm, 0, 1), that means they are blue and their red channel is set depending on their cluster index.

Weirdly, the output is mixed with points that are colored and points that have their original grey value. I haven't had the time to find the bug yet (in clustering? in writer? in acceleration structure?). 

# Logs

Clustering works on 10000 points at the moment and still takes some minutes (I've changed the sigma to get a reasonable number of clusters -> implies more iterations). I took the freedom to added Boost logs in Clustering, Descripting, and Matching to see some progress feedback.

# Bug fixes

- Missing entries in CMakeLists.txt
- Missing object creation of COG in unit test.

Edit: please don't merge yet if you're ok with it. I haven't done a self-review yet.


CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 option(ENABLE_GUI "Build with GUI" ON)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
-set(CMAKE_CXX_FLAGS_DEBUG "-g")
+set(CMAKE_CXX_FLAGS_DEBUG "-g -O3")
```
Fluci: Probably remove again or reduce to O1 or O2: variables can be optimized away, which makes debugging annoying

lib/clustering/mean_shift.cpp
```diff
@@ -33,25 +54,41 @@ std::vector<Cluster> MeanShift::operator()(const PointCloud& cloud) const {
...
 		// ... iterate until convergence
 		do  {
 			iterations++;
 			// perform one iteration of mean shift
 			prevCenter = currCenters[i];
-			currCenters[i] = iterate_mode(currCenters[i], points);
+            std::vector<PointIndex> locals = oracle.find_in_range(currCenters[i], 2*_window_size);
+            if(locals.empty()){
+                BOOST_LOG_TRIVIAL(warning) << "No points in neighborhood, falling back to brute force.";
+                //currCenters[i] = iterate_mode(currCenters[i], points);
```
Fluci: commented??

lib/generic/point_cloud.cpp
```diff
@@ -65,12 +96,42 @@ const Coordinate& PointCloud::Point::z() const {
...
+const ColorChannel& PointCloud::Point::r(const ColorChannel& _new) {
+    _cloud._r[_index] = _new;
+    return _cloud._r[_index];
+}
+
+const ColorChannel& PointCloud::Point::r() const {
+    return _cloud._r[_index];
+}
+
+const ColorChannel& PointCloud::Point::g(const ColorChannel& _new) {
```
Fluci, line 108: keep "void" return type for the moment?

lib/generic/point_cloud.cpp
```diff
@@ -65,12 +96,42 @@ const Coordinate& PointCloud::Point::z() const {
...
+
+const ColorChannel& PointCloud::Point::b() const {
+    return _cloud._b[_index];
+}
+
+ColorChannel PointCloud::Point::intensity() const {
+    return (_cloud._r[_index] + _cloud._g[_index] + _cloud._b[_index])/3.0;
+}
+
+ColorChannel PointCloud::Point::intensity(const ColorChannel& _new) {
```
Fluci, line 130: comment in header missing

lib/spatial/brute_force.h
```diff
@@ -0,0 +1,74 @@
...
+/// Expects points to be Eigen col vectors of size Dim x 1
+/// The point list is a Dim x #Points eigen matrix
+
+template <typename _Precision, int _Dim>
+class BruteForce : public SpatialOracle<Eigen::Matrix<_Precision, _Dim, Eigen::Dynamic, Eigen::RowMajor + Eigen::AutoAlign>, Eigen::Matrix<_Precision, _Dim, 1>, _Precision> {
+public:
+    typedef Eigen::Matrix<_Precision, _Dim, Eigen::Dynamic, Eigen::RowMajor + Eigen::AutoAlign> PointList;
+    typedef Eigen::Matrix<_Precision, _Dim, 1> Point;
+private:
+    PointList _points;
```
Fluci, line 24: This is inefficient: Can we split the class to a safe and an unsafe version? the unsafe version would only hold a pointer to the data, lifetime management is responsibility of the client. The safe version would hold the data and also an instance of the unsafe version who's pointer is set to the internal copy. (maybe split to three classes to safe extra pointer in safe class?)

luhug (approved):

lib/generic/point_cloud.cpp
Fluci: done

lib/point_cloud_filtering/subsample.h
```diff
@@ -0,0 +1,21 @@
...
+#include "point_cloud_filtering.h"
+
+namespace MouseTrack {
+
+class SubSample : public PointCloudFiltering {
+public:
+    SubSample() = default;
+    SubSample(int desiredMaxPoints);
+    virtual ~SubSample() = default;
+    virtual PointCloud operator()(const PointCloud& inCloud) const;
```
Fluci, line 16: accessors to maxPoints

lib/spatial/brute_force.h
Fluci: push to optimization phase

lib/generic/point_cloud.cpp
Fluci: push to optimization phase where we should think about interfaces again: improve interfaces for our use cases

itko (approved):
A little bit overwhelming, but from our discussions and my inspection it seems ok. RGB is a little different from what I thought, but seems ok as well.

PR #26, opened 2018-04-15
Refactoring of Pipeline Factory

Fluci:
As a preparation for the upcoming optimization phase, I've refactored the Pipeline factory.

As a proof of concept, I wrote an "optimized" versions of the MatlabReader and the DisparityRegistration. I want to keep the sequential ones as reference, so we always have a working version we can fall back to.

This pull request demonstrates, how to write pipeline modules and change the configuration via command line.

Example: By default, the new optimized versions are used. Assume you encounter a bug in the registration module and want to fall back to the sequential one. You can do this by adding this command line argument:

```
mousetrack --pipeline-registration=disparity
```

See `--help` for possible options.

At the moment, everything is still built on top of boost's program_options. This is fine for the command line controller, but might be unsuitable for the interactive qt controller. @luhug : We still need to discuss this issue at some point.


CMakeLists.txt
```diff
@@ -10,6 +10,23 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 # options
 option(ENABLE_GUI "Build with GUI" ON)
+option(ENABLE_OPENMP "Build with OpenMP" ON)
```
Fluci: Add note to README.

lib/registration/disparity_registration_simple.cpp
```diff
@@ -0,0 +1,81 @@
...
+/// Maintainer: Felice Serena
+///
+
+#include "disparity_registration_simple.h"
+#include <boost/log/trivial.hpp>
+
+namespace MouseTrack {
+
+PointCloud DisparityRegistrationSimple::
+operator()(const FrameWindow &window) const {
```
Fluci: Think about convention? Where to put reference: maybe as default on parent? Or does the parent provide a "frame" (setting up stuff every child will need)?
Advantages providing default implementation: less files, reference implementation has a "special" position

lib/registration/disparity_registration.cpp
```diff
@@ -20,45 +19,45 @@ PointCloud DisparityRegistration::operator()(const FrameWindow &window) const {
...
     // convert each pixel
-    for (int y = _frame_border - 1; y < disp.rows() - _frame_border; y += 1) {
-      for (int x = _frame_border - 1; x < disp.cols() - _frame_border; x += 1) {
-        double disparity =
-            255 * disp(y, x); // disparity is returned between [0,1], but
-                              // originally stored as [0,255]
-        if (disparity < _min_disparity) {
+    for (int y = border - 1; y < disp.rows() - border; y += 1) {
+      for (int x = border - 1; x < disp.cols() - border; x += 1) {
+        double disparity = 255 * disp(y, x); // disparity is returned between
```
Fluci: comment formatting

luhug (approved):
I don't understand everything 100% but at the moment it looks good.
Is the only difference between normal and optimal versions concurrency?

Fluci:
For the matlabreader: yes, the only difference is that the files get loaded asynchronously (if that's actually an improvement depends highly on the system load and your hardware)

For DisparityRegistration: Not only, the core part is the usage of Eigen matrix operations. They usually are a lot better optimized than what we would like to write (vectorization, prefetching instructions, cache locality, etc.). I still intend to do some improvements, but for that I need to change the implementation of PointCloud (I want to avoid all the resize() and copy operations). But since Registration runs in under 100ms already (on my machine), we should work on something different for the next time.

app/cli_options.cpp
```diff
@@ -19,6 +19,21 @@ op::options_description cli_options() {
...
   ad("first-frame", op::value<int>(), "Desired lowest start frame (inclusive).");
   ad("last-frame", op::value<int>(), "Desired highest end frame (inclusive).");
+  ad("first-stream", op::value<int>(), "Desired lowest start stream (inclusive).");
+  ad("last-stream", op::value<int>(), "Desired highest end stream (inclusive).");
+
+  // pipeline modules
+  ad("pipeline-reader", op::value<std::string>()->default_value("matlab-concurrent"), "Which reader module to use. Valid values: none, matlab, matlab-concurrent");
+  ad("pipeline-registration", op::value<std::string>()->default_value("disparity-cpu-optimized"), "Which registration module to use. Valid values: none, disparity, disparity-cpu-optimized");
+  ad("pipeline-point-cloud-filtering", op::value<std::string>()->default_value("none"), "Which filtering modules to use. Valid values: none, subsample");
+  ad("pipeline-clustering", op::value<std::string>()->default_value("mean-shift"), "Which clustering module to use. Valid values: none, mean-shift");
```
itko, line 29: I agree this helps us in the long run, but is this really important for us right now?

app/pipeline_factory.cpp
```diff
@@ -8,57 +8,140 @@
...
+      getTrajectoryBuilder(options)};
+
   return Pipeline(std::move(reader), std::move(registration),
                   std::move(cloudFiltering), std::move(clustering),
                   std::move(descripting), std::move(matching),
                   std::move(trajectoryBuilder));
 }
 
+std::unique_ptr<Reader>
+PipelineFactory::getReader(const op::variables_map &options) const {
```
itko, line 48: A little concerned with the lack of comments. It could be that this is really standard and obvious stuff, so if you feel there is no need for comments, then no worries.

itko (commented):
1. Some functions comepletely lack comments. I didn't read thoroughly to make sure I understand every single step, but just think twice to make sure this is all understandable.
1. I think we are getting ahead of ourselves in implementing more and more "nice to have" architecture. CLI options such as "getClustering" or "getDescripting" are definitely helpful, but we could have done without them for now. Not only that, I'd argue that it makes for a program that is easier to use, but code that is harder to modify and debug.

That being said, I still think this is good and ready to be merged.

app/cli_options.cpp
Fluci: I'm using this options really hard at the moment (I'm implementing different filters, modifying them easily is a must). I may remind you, that we are now in the improvement phase where we need to be flexible regarding algorithm choice.

app/pipeline_factory.cpp
Fluci: I'll look over it

itko (approved):

PR #28, opened 2018-04-16
Added pipeline Delegate

Fluci:
# Delegate

Since we have quite some parameters to tweak, we might want to process the same frame repeatedly in an interactive fashion using the gui. This might require fine grained control. 

A delegate is an object that is called by another object (our pipeline) and asked different questions. For example: "which frame should the pipeline process next?" or "is there a frame remaining we should process?". Hence a delegate is a very suitable candidate to implement this desired fine grained control.

This will also allow us in the future to skip steps and add caching.

If no delegate is set, the Reader is used directly to acquire frames. Future methods probably should provide simple default implementations.

# Clean up

I noticed, there's an old typedef "FrameIndex" hanging around, which we didn't remove after a merge. I deprecated it and cleaned the current code.

### DIFF
--- a/app/pipeline.cpp
+++ b/app/pipeline.cpp
@@ -136,6 +136,10 @@ void Pipeline::_join() {
 
 void Pipeline::setDelegate(PipelineDelegate *delegate) {
   std::lock_guard<std::mutex> lock(_observer_mutex);
+  if (delegate == nullptr && !_controller_terminated) {
+    throw "It is invalid to set the delegate to nullptr while the pipeline is "
+          "running";
+  }
   _delegate = delegate;
 }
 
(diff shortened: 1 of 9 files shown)


app/pipeline.cpp
```diff
@@ -169,11 +177,23 @@ void Pipeline::runPipeline() {
 
   _clusterChains.clear();
 
-  for (FrameIndex f = _reader->beginFrame(); f < _reader->endFrame(); f += 1) {
-    if (terminateEarly()) {
-      return;
+  if (_delegate != nullptr) {
+    // we have a delegate, use it
+    while (_delegate->hasNextFrame()) {
```
Fluci: lock?

app/pipeline.cpp
```diff
@@ -131,6 +132,17 @@ void Pipeline::_join() {
   _controller_terminated = true;
 }
 
+// delegate handling
+
+void Pipeline::setDelegate(PipelineDelegate *delegate) {
+  std::lock_guard<std::mutex> lock(_observer_mutex);
+  if (delegate == nullptr && !_controller_terminated) {
```
Fluci: probably still not thread safe

luhug (approved):

itko (commented):
Same as #26 . Let's make sure everything we are doing is absolutely necessary. For a 3 person project, I think our architecture is overly complex.

Again, this is done and there's no point reversing it, so it's ready to merge. But I'd like to discuss the balance between architecture and "hacky code". I think you may be able to convince me!

itko (approved):

PR #30, opened 2018-04-16
First implementation of coloring scheme

itko:
This is very basic right now, we can expand functionality later. The important part is that we have a way to randomly color clusters. I tried to make this as high contrast as possible. That is, if we get a color 20 times, all the colors should in theory be quite different from each other.

### DIFF
--- a/lib/color/color.cpp
+++ b/lib/color/color.cpp
@@ -1,0 +1,52 @@
+/** \file
+ * color.cpp
+ *
+ *  Created on: Apr 15, 2018
+ *      Author: itto
+ */
+
+#include "color.h"
+#include <opencv2/contrib/contrib.hpp>
+#include <time.h>
+
+using namespace cv;
+
+namespace MouseTrack {
+
+std::vector<double> GenerateRandomColor() {
+  // Random variable
+  // TODO add parameter for number of clusters that will determine how many
+  // colors to output
+  srand(time(NULL));
+  int i = rand() % 64;
+  // Now multiply to get the intensity
+  int intensity = i * 4;
+  // Create dummy image with the intensity and dummy 0 and full intensity
+  int data[3] = {intensity, 0, 255};
+  Mat dummy(1, 3, CV_8UC1, data);
+  // Dummy output image
+  Mat coloured;
+  // Apply colour map on dummy
+  applyColorMap(dummy, coloured, COLORMAP_JET);
+  // Get the colour from the new image
+  auto red = coloured.at<Vec3b>(0, 0)[2];
+  auto green = coloured.at<Vec3b>(0, 0)[1];
+  auto blue = coloured.at<Vec3b>(0, 0)[0];
+
+  return std::vector<double>{red, green, blue};
+}
+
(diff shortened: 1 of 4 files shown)


lib/color/color.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+using namespace cv;
+
+namespace MouseTrack {
+
+std::vector<double> GenerateRandomColor() {
+  // Random variable
+  // TODO add parameter for number of clusters that will determine how many
+  // colors to output
+  srand(time(NULL));
```
Fluci: To color clusters, I need a way to deterministically get a color for a certain cluster. Could we change the function signature to `GenerateRandomColor(int clusterIndex, int totalClusters)`? Maybe you can use `clusterIndex` and `totalClsuters` as seeds for srand? I could also hand in a seed `GenerateRandomColor(int seed)`. Otherwise seems ok.

lib/color/color.cpp
```diff
@@ -0,0 +1,52 @@
...
+  applyColorMap(dummy, coloured, COLORMAP_JET);
+  // Get the colour from the new image
+  auto red = coloured.at<Vec3b>(0, 0)[2];
+  auto green = coloured.at<Vec3b>(0, 0)[1];
+  auto blue = coloured.at<Vec3b>(0, 0)[0];
+
+  return std::vector<double>{red, green, blue};
+}
+
+std::vector<double> ColorizePointCloud(PointCloud &pc) {
```
Fluci: I think, this should be the responsibility of the client. See https://github.com/itko/mousetrack/blob/master/app/pipeline_writer.cpp#L140

lib/color/color.h
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+#include "generic/point_cloud.h"
+#include <vector>
+
+#pragma once
+
+namespace MouseTrack {
+
+std::vector<double> GenerateRandomColor();
```
Fluci: Very generic, maybe specialize to `ColorKofN(K, N)`? Or at least such that I can pass in the seed `RandomColor(int seed)`.

lib/color/color.cpp
Fluci: Thinking about it, removing the srand() statement should be fine, we don't need to pass in any arguments.

lib/color/color.h
Fluci: nope, I think you've chosen wisely, I changed my mind

Fluci: A doc comment would be nice (chooses random color and returns vector of size 3, Use srand to set seed, etc.).

lib/color/color.cpp
itko: Yes, I felt strange about it in the first place. Removing now.

itko:
@Fluci is this what you meant? I also added functionality for getting N colours. I think it might be useful.
@luhug can you review so we can get visualizations in time

Fluci (approved):
Looks good to me

luhug (approved):

PR #31, opened 2018-04-16
Create build target for "sandbox"

itko:
Following our discussion for a way to quickly play around with code we wrote, I created a "sandbox" for myself where I can quickly build and debug code I develop. We can add ones for you as well if you want.

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -83,6 +83,11 @@ target_link_libraries(appTests
     ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY}
     )
 
+# create the sandbox_itto executable
+add_executable(sandbox_itto sandbox/sandbox_itto.cpp)
+target_link_libraries(sandbox_itto
+    libmousetrackAll
+    )
 
 # create a target for the headers, we don't actually need the result
 # but this way, IDE's also show header files in their project overviews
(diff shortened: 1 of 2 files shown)


Fluci:
Interesting, I expected something like a general sandbox for algorithms (one target per algorithm where we can show in detail how the algorithm behaves), less a personal sandbox (one target per person where every person puts random stuff). 

What is the lifetime of code in this folder? 
Will we have to maintain it?
Or is it a debugging environment? 
What is the difference to a unit test so far? 

I'm very concerned about adding our names as actual executable names, if it's only a personal scratch pad, there's no point actually merging it in master, just keep it as local/separate branch.

itko:
1. This is different from a unit test. This allows me (or whoever wants) to play around with code and debug it and see what's going on. This might involve using multiple modules at once, or running "tests" that are more comprehensive than unit tests.
1. I don't believe there should be one executable per algorithm as you say. My goal is to reduce the uncertainty that I have when pushing code without requiring me to run through the whole pipeline to make sure my code works. This way I can iterate faster and reduce silly bugs.
1. I agree this doesn't belong in master. And if this were a project with 8+ people working on it, I wouldn't put it master. But the convenience of not having to constantly merge this into any branch I'm working on currently outweighs the need to keep it off master. Once we are ready for releases, we can kick this out
1. The code will only be maintained by me and it is completely fine if it loses compatibility. I'm more concerned with the actual cmake structure than the code itself.

Fluci:
If it's in master, everybody needs to maintain it in a sense, because if it fails to compile (resulting from an interface change or so), the entire project is busted. We should discuss this in the meeting today.

PR #33, opened 2018-04-19
Catch exceptions at highest levels

Fluci:
Loosing time because of terrible exception handling is unacceptable, I've added try-catch statements at the most important places so we at least have an idea of what happened. 

Now exception handling is no longer terrible, only bad.


app/pipeline.cpp
```diff
@@ -202,15 +202,37 @@ void Pipeline::runPipeline() {
       if (terminateEarly()) {
         return;
       }
-      processFrame(f);
+      try {
```
Fluci: a new method "processFrameSafe()" might be better than repeating code

luhug (approved):

itko (approved):
Very useful. Thanks!

PR #34, opened 2018-04-19
Feature/fs/filters

Fluci:
# Stackable filters

The pipeline supports fully stackable filters (create multiple filters, put them in a vector, the pipeline pushes the frame windows through each filter sequentially).

You can also stack filters from the command line, just add:

```
--pipeline-frame-window-filtering <filter1> <filter2> ...
```

One caveat: I haven't figured out a descent way to use the same filter (for example disparity-morphology) multiple times with different parameters (once opening, once closing for example) using the command line.

I'm still thinking about this problem, but I think we might have reached a point, where we should use some hierarchical description of the pipeline (an xml config file or so). No point in overdoing the command line interface.

# Filters

I've added following filters based on opencv, all running only on the disparity image:

- Gaussian bluring
- Median smoothing
- Bilateral smoothing
- Morphological openning and closing

I've also implemented a statistical outlier removal algorithm for the 3d point cloud: It looks at a point's k nearest neighbors, calculates their mean and variance, and if the point is within `alpha*sigma`, it is an inlier, otherwise an outlier. 

# Signal handling

The command line controller now handles SIGTERM and SIGINT signals, hitting ^C once in the terminal causes the pipeline to stop as soon as possible (after finishing the running module), proper clean up is performed.

Hitting ^C a second time, causes the application to terminate immediately (chained point clouds can't be written for example).


app/cli_options.cpp
```diff
@@ -24,17 +24,37 @@ op::options_description cli_options() {
...
-  // mean-shift
+  // frame window post processing
+  ad("disparity-gauss-k", op::value<int>()->default_value(3), "Patch diameter in x/y direction, must be positive and integer. k = 0 results in a 1x1 patch size.");
+  ad("disparity-gauss-sigma", op::value<double>()->default_value(0.5), "Gaussian standard deviation in x/y direction, must be positive.");
+  ad("disparity-bilateral-diameter", op::value<int>()->default_value(2), "Patch diameter in x/y direction, must be positive and integer.");
+  ad("disparity-bilateral-sigma-color", op::value<double>()->default_value(0.1), "Filter sigma in color space.");
+  ad("disparity-bilateral-sigma-space", op::value<double>()->default_value(50), "Filter sigma in coordinate space.");
+  ad("disparity-median-diameter", op::value<int>()->default_value(2), "Patch diameter in x/y direction, must be positive and integer.");
+  ad("disparity-morphology-diameter", op::value<int>()->default_value(2), "Patch diameter in x/y direction, must be positive and integer.");
+  ad("disparity-morphology-operation", op::value<std::string>()->default_value("open"), "Specific morphological operation. Valid values: open, close");
```
Fluci: separate into two separate filters? "disparity-morph-open"/"disparity-morph-close"?

app/pipeline_factory.cpp
```diff
@@ -77,6 +91,101 @@ PipelineFactory::getReader(const op::variables_map &options) const {
...
+    double sigmaColor = options["disparity-bilateral-sigma-color"].as<double>();
+    double sigmaSpace = options["disparity-bilateral-sigma-space"].as<double>();
+    ptr->diameter(diameter);
+    ptr->sigmaColor(sigmaColor);
+    ptr->sigmaSpace(sigmaSpace);
+    return ptr;
+  }
+  if (target == "disparity-morphology") {
+    auto ptr = std::make_unique<DisparityMorphology>();
+    int diameter = options["disparity-bilateral-diameter"].as<int>();
```
Fluci: wrong

README.md
```diff
@@ -8,14 +8,14 @@ This project will be awesome!
...
-$ sudo apt install -y cmake libboost-all-dev libopencv-dev libeigen3-dev libpng16-dev
+$ sudo apt install -y cmake libboost-all-dev libopencv-dev libeigen3-dev libpng-dev
 ```
 
 - CMake: `$ sudo apt install cmake`
 - Boost.Test: `$ sudo apt install libboost-all-dev`
 - [OpenCV](https://docs.opencv.org/trunk/d7/d9f/tutorial_linux_install.html): `$ sudo apt install libopencv-dev`
 - Eigen: `$ sudo apt install libeigen3-dev`
-- We need to read PNG files: `$ sudo apt install libpng16-dev`
+- We need to read PNG files: `$ sudo apt install libpng-dev`
```
luhug: Do I need to install the new one or is it compatible with the old dependency?

app/pipeline_factory.cpp
```diff
@@ -77,6 +91,110 @@ PipelineFactory::getReader(const op::variables_map &options) const {
...
+
+std::unique_ptr<FrameWindowFiltering>
+PipelineFactory::getWindowFiltering(const std::string &target,
+                                    const op::variables_map &options) const {
+  if (target == "disparity-gauss") {
+    auto ptr =
+        std::unique_ptr<DisparityGaussianBlur>(new DisparityGaussianBlur());
+    int k = options["disparity-gauss-k"].as<int>();
+    double sigma;
+    if (options.count("disparity-gauss-sigam")) {
```
luhug: typo "disparity-gauss-sigam" -> disparity-gauss-sigma

app/pipeline_factory.cpp
```diff
@@ -77,6 +91,110 @@ PipelineFactory::getReader(const op::variables_map &options) const {
...
+      sigma = options["disparity-gauss-sigma"].as<double>();
+    } else {
+      sigma = 0.3 * k + 0.8;
```
itko: What is this? Explain maybe

luhug (changes requested):

app/pipeline_factory.cpp
Fluci: thanks, those are the tricky ones to track down later :)

app/pipeline_factory.cpp
Fluci: good point, I'll add a comment. In short: they come from OpenCV, I haven't thought about the exact meaning, but I guess it's something like "make sure, 90% of the guassian weight is inside the window". At least that's what `0.3*k` suggests to me. `+0.8` might be some lower bound.

README.md
Fluci: Hmm, looking at it again: I just checked apt-get to make sure I don't break anything, I do. So I'll roll it back. 

That's a change I did one or two weeks ago, OpenCV also uses libpng, and somewhere it is only listed to be `libpng-dev`. So I adjusted our dependency to be compatible (only necessary in the readme). In general, it can be annoying if A uses Bv1 and C, while C uses Bv2. I tried to avoid this, but since it would cause more harm then help, rolling back ...

Note: On travis it's actually running `libpng12-dev`, since 1.6 isn't available. This tells me, that we don't actually use any unstable interfaces, so it doesn't matter too much in the end.

luhug (approved):

itko (approved):

PR #40, opened 2018-04-25
Add notebooks

itko:
Current notebooks for colorizing and masking.

Git is not very good for versioning notebooks, so this is more for us to share results. Perhaps each time we have an interesting results we can:
1. Post pictures, point clouds, etc.
2. If it looks promising then create a pull request and merge

Also, if you add "nice to have" functionality (multiple frames, etc.) then that's also worth merging in so we can all use it.


Fluci (approved):

Fluci:
How do I properly set up my environment to get the notebooks to work? I keep getting the error `ImportError: No module named matplotlib`, I've tried to install matplotlib it via conda, pip, and pip3 but it still fails.

PR #41, opened 2018-04-25
Added frame grabber script

luhug:
The script pulls all files for the specified frames into a new directory.
Arguments must be given in the following order:
Input directory (where the whole bag is extracted)
Output directory
Number of first frame that you want
Step size (if step size is n, it only copies every nth frame)
Number of the last frame that you want.

### DIFF
--- a/scripts/grab_frames.sh
+++ b/scripts/grab_frames.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+# Extracts a subset of frames from input folder into output folder
+# $1: Input folder
+# $2: Output folder
+# $3: Start frame
+# $4: Step
+# $5: End frame
+
+# Checks if file exists, copies if yes, throws waring if not.
+function check_and_copy {
+    if [ -f $1 ]
+then
+  cp $1 $2
+else
+  echo "WARNING:" $1 "not found"
+fi
+}
+
+# Check correct number of arguments
+if [ $# -lt 5 ]
+then
+  echo "Usage: $0 INPUT_DIR OUTPUT_DIR START_FRAME STEP END_FRAME"
+  exit 1
+fi
+
+# If output directory doesn't exist, create it.
+if [ ! -d $2 ]
+then
+   mkdir $2
+fi
+
+check_and_copy $1/params_R.csv $2/
+check_and_copy $1/params_camchain.csv $2/
+
+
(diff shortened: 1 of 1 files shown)


scripts/grab_frames.sh
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+check_and_copy $1/params_R.csv $2/
+check_and_copy $1/params_camchain.csv $2/
+
+
+for i in `seq $3 $4 $5`;
+do
+    check_and_copy $1/params_f_$i.csv $2/
+    for j in `seq 1 4`;
```
Fluci: careful about the hard coded assumption that we have 4 streams, there are also bags with only the first two streams because cameras didn't work, but probably ok for most cases

scripts/grab_frames.sh
```diff
@@ -0,0 +1,45 @@
...
+if [ ! -d $2 ]
+then
+   mkdir $2
+fi
+
+check_and_copy $1/params_R.csv $2/
+check_and_copy $1/params_camchain.csv $2/
+
+
+for i in `seq $3 $4 $5`;
```
Fluci: very elegant

scripts/grab_frames.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+# Extracts a subset of frames from input folder into output folder
+# $1: Input folder
+# $2: Output folder
+# $3: Start frame
+# $4: Step
+# $5: End frame
+
+# Checks if file exists, copies if yes, throws waring if not.
```
Fluci: typo: warning

Fluci (approved):
Just some thoughts, fine to merge in my opinion.

Fluci:
One other thought: Do you think it would make sense to put every collection of scripts in its own subdirectory? This might avoid some confusion in the future, I didn't think about that when I added the extraction script.

Fluci:
I just run the script: 

linux needs adjusted permissions: chmod +x ./grab_frames.h
linux has trouble with the Windows line ending characters (/bin/bash^M: bad interpreter), look at [this](https://askubuntu.com/questions/304999/not-able-to-execute-a-sh-file-bin-bashm-bad-interpreter) for fixes I you encounter the issue.

itko (approved):
Did not check. Automatic approval.

PR #42, opened 2018-04-28
Flann, kmeans, mean-shift-cpu-optimized

Fluci:
In short:

- Added Flann as SpatialOracle.
- Oracle Factory
- created first draft of a cpu optimized mean shift
- create draft for kmeans
- bug fixes

# FLANN

I hopped, FLANN would be good enough for our queries (since it only returns approximate answers).

My test runs suggest, that FLANN is "good enough" for kmeans. But it really struggles to converge points for mean-shift, where UniformGrid works solid.

By now, there are a lot of parameters around to tweak: FLANN is a collection of algorithms, most of them are slower or work really bad for us. I've decided to use a K-means index with default settings for the moment.

# Oracle Factory

An alternative to a factory would have been to set an oracle at the start as a sub-object of a module. I've decided against it:

- It is not clear, what parameters should be handed in without looking at the data (knowing the bounding box of the data or other characteristics might be useful).
- We might decide to use multiple oracles based on different requirements (example for mean shift: it could use flann to get a rough convergence for the points and then switch to the more costly UniformGrid for fine tuning).

Disadvantage of this approach: General structures like the Neighborhood of UniformGrid need to be built for each frame, this shows to be expensive for example for kmeans. I'd still prefer to add a caching based solution in the future instead of explicitly setting oracles in the PipelineFactory.

# Optimized mean shift

I refactored the original mean shift by adding a "convergePoints" and "mergePoints" method which I then improved independently. There was not a lot to do for the convergence part: I just formulated everything as Eigen operations (we copied the values anyway, so we can as well use optimized library functions than writing our own loop). The merging part now also uses a spatial oracle, hence it has a smaller big-O complexity. 

Measurements showed, that the optimized mean shift indeed is faster. (example: 20'000 points, sigma:0.02,merge-threshold:0.01, BruteForce: 60s unoptimized vs. 11s optimized)

Mean-shift is not yet parallelized with OpenMP. I first have to investigate if FLANN supports this.

# Kmeans

I created a first rough draft for KMeans (the initialization could definitely be improved) to get a feeling how much better KMeans performs. 

First results are very interesting: There's no clear winner.

If you use UniformGrid, mean shift is faster. But using Flann gives kmeans the edge regarding speed (UniformGrid has a huge start up time constructing the neighborhood information: A closed form solution would probably solve the problem, but I couldn't figure one out for N dimension until now).

Mean-shift seems to create a lot more "noise" clusters. Visual inspection shows, that both algorithms seem to find the mouse cluster (Running on around 20'000 subsampled points). Only exception seems to be when using mean-shift together with FLANN: FLANN seems to return far to approximate answers as the points would converge reasonably.

# Bug fixes

Some minor things like typos or forgotten `const`'s

# Edit

I took a look at caching the oracle and implemented it. I'm still looking for a thread_local solution, but for the moment it gives the Kmeans/UniformGrid combo a very good speed up (though, FLANN is still faster by a factor of 2-3).

I slightly improved the convergence heuristic for kmeans.


Fluci:
FLANN crashes when loop is parallelized with openmp -> get mutex within in flann wrapper as momentary work around

luhug (approved):
I definitely don't understand everything, but I couldn't find any errors for the moment.

PR #43, opened 2018-04-30
Feature/fs/robust reader

Fluci:
As discussed, I've adjusted the MatlabReader such that it skips frames that aren't around/complete (no warning printed ... desirable?). For this, a Reader now supports the same `hasNextFrame()/nextFrame()` mechanism as the pipelineDelegate. I've moved the declarations into a new `FrameIterable` interface (I hope to be able to unify the calls one day and get rid of the branching within the pipeline ...).

I've also changed the extraction script such that the raw disparity map no longer is exported, we don't use it anyway (you need to add the corresponding value in your config of the script). I've also moved the preallocation stuff inside the for loop by default. This way we are more flexible to play around with parallelization of the extraction.


luhug (approved):
Looks fine to me.

itko (approved):

PR #44, opened 2018-05-05
Added background subtraction filter module

luhug:
# How to use:
* Add `background-subtraction` as argument of `--pipeline-frame-window-filtering`
* Specify path to empty cage frame with `--background-subtraction-cage-directory`. When none is specified, it uses the `-src-dir`.
* Specify which frame to use with `--background-subtraction-cage-frame`. Default is 1.

# Usage example:
`./app/mousetrack -c -s ../data/2Lasers/ -l debug --first-frame 700 --last-frame 700 --first-stream=1 --last-stream=4 --pipeline-timer --pipeline-clustering=single-cluster --pipeline-frame-window-filtering background-subtraction --background-subtraction-cage-frame 1 --background-subtraction-cage-directory ../data/2Lasers_empty -o ../data/output`

Note: I also added VSCode support to .gitignore, because I wanted to try it out.


app/cli_options.cpp
```diff
@@ -45,6 +45,9 @@ op::options_description cli_options() {
   ad("disparity-morph-open-shape", op::value<std::string>()->default_value("rect"), "Shape of kernel for morphological operation. Valid values: rect, ellipse, cross");
   ad("disparity-morph-close-diameter", op::value<int>()->default_value(2), "Patch diameter in x/y direction, must be positive and integer.");
   ad("disparity-morph-close-shape", op::value<std::string>()->default_value("rect"), "Shape of kernel for morphological operation. Valid values: rect, ellipse, cross");
+  ad("background-subtraction-threshold", op::value<double>()->default_value(0.01), "Threshold for background subtraction.");
```
Fluci: That's for the color range [0,1]? Might be worth to put a note to avoid confusion.

app/cli_options.cpp
```diff
@@ -45,6 +45,9 @@ op::options_description cli_options() {
...
+  ad("background-subtraction-cage-frame", op::value<int>()->default_value(1), "Number of a frame with empty cage");
+  ad("background-subtraction-cage-directory", op::value<std::string>()->default_value(""), "Path to directory with empty cage. Default: src-dir");
```
Fluci: Might be cleaner to have no default value and do a check with `options.count("background-subtraction-cage-directory")>0`, but that's very much an edge case so do as you please.

app/pipeline_factory.cpp
```diff
@@ -193,6 +194,23 @@ PipelineFactory::getWindowFiltering(const std::string &target,
...
   }
+  if (target == "background-subtraction") {
+    std::string path;
+    if (options["background-subtraction-cage-directory"].as<std::string>() == "") {
+      path = options["src-dir"].as<std::string>();
+    } else {
+      path = options["background-subtraction-cage-directory"].as<std::string>();
+    }
+    MatlabReader reader (path);
+    reader.setBeginFrame(options["background-subtraction-cage-frame"].as<int>());
```
Fluci: you don't need to set these two frame-lines, the values are set by the preflight in the constructor, these methods only shrink the valid range for future queries. But since you don't hand off the reader to other parts of the app, there's no point in shrinking the range.

app/pipeline_factory.cpp
```diff
@@ -193,6 +194,23 @@ PipelineFactory::getWindowFiltering(const std::string &target,
...
+    reader.setEndFrame(options["background-subtraction-cage-frame"].as<int>());
+    reader.setBeginStream(options["first-stream"].as<int>());
```
Fluci: Very interesting, probably the right choice, I didn't expect to see this. :+1:

lib/frame_window_filtering/background_subtraction.cpp
```diff
@@ -0,0 +1,89 @@
...
+  // Cycle through all streams
+  for (size_t i = 0; i < _cage_frame.frames().size(); i++) {
+    // Copy the stuff we need from the FrameWindow objects
+    Picture cage_image = _cage_frame.frames()[i].referencePicture;
+    Picture mouse_image = window.frames()[i].referencePicture;
+
+    // ALL the dimensions must match
+    if (!(cage_image.rows() == mouse_image.rows() &&
+          cage_image.cols() == mouse_image.cols())) {
+      // If it fails, return the input (do nothing)
```
Fluci: This is probably worth at least an "info" log since it's still quite unexpected to the user.

lib/frame_window_filtering/background_subtraction.cpp
```diff
@@ -0,0 +1,89 @@
...
+  // empty
+}
+
+FrameWindow BackgroundSubtraction::operator()(const FrameWindow &window) const {
+    // Initializing return object
+  std::vector<Frame> frames(window.frames().size());
+  // Cycle through all streams
+  for (size_t i = 0; i < _cage_frame.frames().size(); i++) {
+    // Copy the stuff we need from the FrameWindow objects
+    Picture cage_image = _cage_frame.frames()[i].referencePicture;
```
Fluci: This creates a copy of the image, you might want to switch it to a constant reference `const Picture& carge_image` to omit the copy operation, same for next line.

lib/frame_window_filtering/background_subtraction.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    // Convert back to Eigen
+    Eigen::MatrixXd mask;
+    cv::cv2eigen(maskcv, mask);
+    mask = mask / 255;
+
+    // A very low threshold means there's no significant bright
+    // spots, i.e. no mouse => set mask to zeros
+    if (thresh_otsu < _threshold) {
+      mask *= 0;
```
Fluci: `mask.setZero()`?

lib/frame_window_filtering/background_subtraction.cpp
```diff
@@ -0,0 +1,89 @@
...
+const FrameWindow &BackgroundSubtraction::cage_frame() const {
+  return _cage_frame;
+}
+
+void BackgroundSubtraction::cage_frame(FrameWindow &cage_frame) {
+
+  _cage_frame = cage_frame;
+}
+
+const double BackgroundSubtraction::threshold() const { return _threshold; }
```
Fluci: Since you don't return a reference, there's no point in making the return type constant, `double BackgroundSubtraction::threshold() const {...}` is enough.

lib/frame_window_filtering/background_subtraction.cpp
```diff
@@ -0,0 +1,89 @@
...
+    mask = mask / 255;
+
+    // A very low threshold means there's no significant bright
+    // spots, i.e. no mouse => set mask to zeros
+    if (thresh_otsu < _threshold) {
+      mask *= 0;
+    }
+
+    // Build Frame object...
+    Frame frame = window.frames()[i];
```
Fluci: I'm not sure about the entire copy action down here ... it feels cleaner to copy everything at the start like `newWindow = window` and then to just update the modified values. This would avoid bugs in the future where additional properties are added to FrameWindow. I actually haven't thought about this until now, I have to check my filters regarding this issue, too.

Fluci (commented):
Looks mostly ok to me, just some details.

app/cli_options.cpp
luhug: You're 100% right, I didn't consider this option.

app/cli_options.cpp
luhug: done

itko (approved):

PR #45, opened 2018-05-06
Feature/fs/ros bag reader

Fluci:
One more fairly large pull request.

# RosBagReader

Allows you to read bag files directly. For this, specify `-s <pathToBagFile>` and `-y <pathToYamlCamChainFile>` on the command line. It doesn't support all features, as one might like (for example first/last-stream) but it works good enough to avoid the Matlab script in the future. 

To compile, make sure following packages are installed on your system: 

``` bash
sudo apt install librosbag-dev libsensor-msgs-dev libimage-transport-dev libyaml-cpp-dev libbz2-dev
``` 

If you have trouble, getting it to compile, you can turn off the entire RosBag module by handing cmake a flag:

```
cmake -DENABLE_ROSBAG=NO ..
```

There are still some TODO's in the MatlabReader, but I don't see me fixing them in the next weeks. They aren't really crucial for us.

# Refactoring

There's also some refactoring going on, for example: 

- The command line argument `src-dir` is now called `src` (since it also takes files). 

- Images are now explicitly name `PictureI` and `PictureD`, instead of using Eigen's MatrixXd/MatrixXi. This is more flexible and clearly defines which convention is used for the intensities ([0,1] for double and [0,255] for unsigned chars). It also makes matrices row-major by default which seems to be the convention of other picture libraries, hence we can interface with them more efficiently.

- I've also changed the DisparityMap, it's no longer a separate class but only a typedef of an Eigen Matrix (zMap() was the wrong name anyway and it seems to make more sense to treat a disparity map just like any other kind of image).

# Matlab script

As preparation for the RosBagReader, I've refactored the undistrect function, there's no no more redundant code and I removed everything we don't need.

# Unit tests

The unit test look a bit messy, but that's mostly clang's auto-formatting, I only changed the above mentioned refactoring issues.

# Pipeline Writer

To verify my implementation, I've also added the export of FrameWindows in the pipelineWriter. By default, they are still turned off (hard-coded). I would like to make it configurable via the command line, but I think there are more urgent issues at the moment. But it's mostly a matter of adjusting `main.cpp` and `cli_options.cpp` if somebody wants to do it him-self.

# Bug fixing

As always, I fixed some bugs I encountered (for example missing proper termination in `read_png.cpp`)

# Edit

I nearly forgot, it builds upon #43, just ignore this fact, I'll take care of it when merging.


CMakeLists.txt
```diff
@@ -11,9 +11,11 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 # options
 option(ENABLE_GUI "Build with GUI" ON)
 option(ENABLE_OPENMP "Build with OpenMP" ON)
+option(ENABLE_ROSBAG "Build with Ros-Bag support" ON)
 
 message("GUI: " ${ENABLE_GUI})
 message("OpenMP: " ${ENABLE_OPENMP})
+message("OpenMP: " ${ENABLE_ROSBAG})
```
Fluci: typo

app/CMakeLists.txt
```diff
@@ -41,6 +46,9 @@ find_package( Boost COMPONENTS
 
 ADD_DEFINITIONS(-DBOOST_LOG_DYN_LINK)
 
+find_package( yaml-cpp REQUIRED )
```
Fluci: move to ENABLE_ROSBAG region

app/ros_bag_reader.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+  cpath = resolve_symlink(cpath.string(), _followSymlinkDepth);
+
+  if (!fs::is_regular_file(cpath)) {
+    BOOST_LOG_TRIVIAL(info)
+        << "Path to camchain file does not point to a file: " << cpath;
+    throw "Camchain file path not a regular file.";
+  }
+
+  _valid = true;
```
Fluci: move at end

app/ros_bag_reader.cpp
```diff
@@ -0,0 +1,385 @@
...
+    frame.referencePicture = im.cast<PictureD::Scalar>() * (1.0 / 255);
+    im = readImageFromMsg(_frameMessages[f][index(s, Disparity)]);
+
+    frame.rawDisparityMap = im.cast<PictureD::Scalar>() * (1.0 / 255);
+    frame.normalizedDisparityMap = normalizeDisparity(im);
+    // camInfo holds: focallength, ccx, ccy
+    const auto camInfo = _frameMessages[f][index(s, CamInfo)]
+                             .instantiate<sensor_msgs::CameraInfo>();
+    if (camInfo == nullptr) {
+      BOOST_LOG_TRIVIAL(warning) << "Couldn't instatiate CamerInfo, setting "
```
Fluci: typo

app/ros_bag_reader.cpp
```diff
@@ -0,0 +1,385 @@
...
+                 [](rosbag::View &v) { return v.begin(); });
+
+  std::vector<rosbag::View::iterator> viewEnds;
+  std::transform(views.begin(), views.end(), std::back_inserter(viewEnds),
+                 [](rosbag::View &v) { return v.end(); });
+
+  BOOST_LOG_TRIVIAL(debug) << "Collecting bag handles.";
+  // Collect handles to all relevant messages
+  while (!reachedEnd(viewIters, viewEnds)) {
+    logIters(viewIters);
```
Fluci: op

app/ros_bag_reader.cpp
```diff
@@ -0,0 +1,385 @@
...
+    const std::vector<rosbag::View::iterator> &ends) const {
+  // get common time
+  ros::Time highestTime;
+  for (const auto &v : iters) {
+    highestTime = std::max(v->getTime(), highestTime);
+  }
+
+  // for each stream, find closest value to highestTime
+  for (size_t i = 0; i < iters.size(); ++i) {
+    auto &it = iters[i];
```
Fluci: might a copy fix the issue?

app/ros_bag_reader.h
```diff
@@ -0,0 +1,99 @@
...
+#include "types.h"
+
+#include <rosbag/message_instance.h>
+#include <rosbag/view.h>
+
+#include <yaml-cpp/yaml.h>
+
+namespace MouseTrack {
+
+class RosBagReader : public Reader {
```
Fluci: doc comment

lib/generic/read_png.cpp
```diff
@@ -40,12 +40,16 @@ Eigen::MatrixXi read_png(const std::string &path) {
 
   png_infop info_ptr = png_create_info_struct(png_ptr);
   if (!info_ptr) {
+    png_destroy_read_struct(&png_ptr, (png_infopp)NULL, (png_infopp)NULL);
```
Fluci, line 46: fclose

Fluci:
Add rosbag apt source to readme

itko (approved):

Fluci:
Added changes to work with background-subtraction that is now in master.

luhug (approved):

PR #46, opened 2018-05-16
Feature/fs/gui

Fluci:
As preparation for the point cloud view. It looks very large, but that's mostly from the merge with master but take a look at your gui files and tell me if they're ok.

- Updated gui code such that it merges with master.
- Removed some memory leaks.
- Clang-formatted stuff ....


Fluci:
Need some more time to think about ownership in qt.

PR #48, opened 2018-05-17
Validation

luhug:
# A python script for validation
Caution: it has not been tested with multiple control points yet, but I see no reason why it shouldn't work.

You have to give it three arguments.
1) Where your controlPoints.csv's are (output folder of ./app/mousetrack)
2) Where your png's and param.csv's are (input folder of ./app/mousetrack)
3) The output folder for the python script

Then it will mark the control points on the corresponding pngs and save the result in the output directory.

You can also set an option in the script to get a preview of the validation images.

Also, I fixed an issue where the application would crash if there's no raw disparity map.


scripts/validation.py
```diff
@@ -0,0 +1,143 @@
...
+    return np.asarray(matrix,dtype='float64')
+
+def project_point(control_point, projection_matrix, params):
+    focallength = float(params[0])
+    ccx = float(params[2])
+    ccy = float(params[3])
+    cp_hom = np.row_stack((control_point,[[1]]))
+    p4D = np.matmul(projection_matrix , cp_hom)
+
+    p2D = np.array([p4D[0]/p4D[2]*focallength, p4D[1]/p4D[2]*focallength])
```
Fluci, line 62: This line fails, if a controlPoint at (0,0,0) is passed. Maybe ignore them further up? Or should a clean controlPoints.csv never contain a point at (0,0,0)? since the value doesn't make sense, I use it as a placeholder to show, this point should be ignored. We might need to adjust the main exectuable.

Fluci (approved):
Little issue with line 62, otherwise looks good.

itko (approved):

PR #51, opened 2018-05-28
Feature/fs/hog

Fluci:


app/pipeline_writer.cpp
```diff
@@ -77,14 +95,159 @@ void PipelineWriter::newFrameWindow(FrameNumber f,
...
+    fs::create_directory(base);
+  }
+  Eigen::MatrixXd params(window->frames().size(), 4);
+  size_t maxLabelCount = 6;
+  for (StreamNumber s = 0; (size_t)s < window->frames().size(); ++s) {
+    maxLabelCount = std::max(maxLabelCount, window->frames()[s].labels.size());
+  }
+  auto colors = nColors(maxLabelCount);
+  // fix the first 6 ones so we can decode them
+  colors[0] = std::vector<double>{255, 0, 0};
```
Fluci: don't like hard-coding

app/pipeline_writer.cpp
```diff
@@ -77,14 +95,159 @@ void PipelineWriter::newFrameWindow(FrameNumber f,
...
+    // clang-format on
+    const Frame &frame = window->frames()[s];
+
+    writePng(frame.referencePicture, ref.string());
+    writePng(frame.rawDisparityMap, dispRaw.string());
+    writePng(frame.normalizedDisparityMap, dispNorm.string());
+
+    for (size_t l = 0; l < frame.labels.size(); ++l) {
+      const auto &label = frame.labels[l];
+      fs::path labelP =
```
Fluci: formatting

app/pipeline_writer.cpp
```diff
@@ -77,14 +95,159 @@ void PipelineWriter::newFrameWindow(FrameNumber f,
...
+              l);
+      writePng(label, labelP.string());
+    }
+    // write max-label image
+    if (!frame.labels.empty()) {
+      int rows = frame.labels[0].rows();
+      int cols = frame.labels[0].cols();
+
+      // ignore background
+      std::set<size_t> ignore{5};
```
Fluci: Remove hard-coding

app/pipeline_writer.cpp
```diff
@@ -77,14 +95,159 @@ void PipelineWriter::newFrameWindow(FrameNumber f,
...
+              maxL = l;
+            }
+          }
+          allLabels.at<cv::Vec3d>(y, x)[0] = colors[maxL][0];
+          allLabels.at<cv::Vec3d>(y, x)[1] = colors[maxL][1];
+          allLabels.at<cv::Vec3d>(y, x)[2] = colors[maxL][2];
+        }
+      }
+      fs::path allPath =
+          base /
```
Fluci: formatting

app/pipeline_writer.cpp
```diff
@@ -77,14 +95,159 @@ void PipelineWriter::newFrameWindow(FrameNumber f,
   write_csv(paramsPath.string(), paramsVec);
 }
 
+void PipelineWriter::newFilteredFrameWindow(
+    FrameNumber f, std::shared_ptr<const FrameWindow> window) {
+  if (!writeFilteredFrameWindow) {
```
Fluci: Flags to turn on/off sections would be nice

lib/classifier/knn.cpp
```diff
@@ -0,0 +1,52 @@
...
+  auto max = _X_train.colwise().maxCoeff();
+  Eigen::VectorXd size = max - min;
+  double maxR = size.norm();
+  // heuristic
+  double cellWidth = size.minCoeff() / 5;
+  int dims = X_train.cols();
+  // don't use Uniform grid, it uses memory proportional to 2^dimensions
+  //_oracle = std::make_unique<UniformGrid<double, -1>>(maxR, cellWidth, dims);
+  //_oracle = std::make_unique<BruteForce<double, -1>>();
+  _oracle = std::make_unique<Flann<double, -1>>();
```
Fluci: Use factory

lib/classifier/knn.cpp
```diff
@@ -0,0 +1,52 @@
...
+Eigen::MatrixXd KnnClassifier::predict(const Mat &X_test) const {
+  Eigen::MatrixXd result;
+  result.setZero(_highestLabel + 1, X_test.cols());
+  for (int i = 0; i < X_test.cols(); ++i) {
+    auto closest = _oracle->find_closest(X_test.col(i), _k);
+    for (auto c : closest) {
+      int l = _y_train[c];
+      result(l, i) += 1;
+    }
+    result.col(i) /= (result.col(i).sum() + 0.000001);
```
Fluci: figure out convention: should it be normalized? or is this responsibility of the client?

lib/classifier/knn.h
```diff
@@ -0,0 +1,35 @@
...
+  int k() const;
+  void k(int newK);
+
+private:
+  std::unique_ptr<Oracle> _oracle;
+  Mat _X_train;
+  Vec _y_train;
+
+  // number of nearest neighbors to consider
+  int _k = 3;
```
Fluci: set k higher by default? 9, 11?

lib/frame_window_filtering/background_subtraction.cpp
```diff
@@ -63,21 +63,13 @@ FrameWindow BackgroundSubtraction::operator()(const FrameWindow &window) const {
     // spots, i.e. no mouse => set mask to zeros
     if (thresh_otsu < _threshold * 255) {
       mask.setZero();
```
Fluci, line 65: Does `setZero` need arguments??

lib/frame_window_filtering/hog_labeling.cpp
```diff
@@ -0,0 +1,152 @@
...
+  // vectors
+  cv::HOGDescriptor hog(windowSize, blockSize, blockStride, cellSize, _nbins);
+
+  // compute settings
+  cv::Size windowStride;
+  cv::Size padding;
+
+  // build locations for sliding window
+  const auto &first = result.frames()[0];
+  std::vector<cv::Point> locations = slidingWindows(
```
Fluci, line 75: refactor such that an other strategy could easily be used (wrap in HogCandidateDetector object or so)

lib/frame_window_filtering/hog_labeling.cpp
```diff
@@ -0,0 +1,152 @@
...
+          // distribute labels
+          for (int l = 0; l < _numLabels; ++l) {
+            frame.labels[l](y, x) += labels(l, w);
+          }
+        }
+      }
+    }
+    // normalize label weights
+    // make sure the highest value is 1 and the lowest value is 0
+    bool normalizePeaks = true;
```
Fluci: Things are very messy down here, think again about them

lib/frame_window_filtering/hog_labeling.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+  // hog settings
+  cv::Size windowSize(_windowWidth, _windowHeight);
+  cv::Size blockSize(_blockWidth, _blockHeight);
+  cv::Size blockStride(_blockStrideWidth, _blockStrideHeight);
+  cv::Size cellSize(_cellWidth, _cellHeight);
+
+  // create hog descriptor: it is able to tansform images patches to feature
+  // vectors
+  cv::HOGDescriptor hog(windowSize, blockSize, blockStride, cellSize, _nbins);
```
Fluci, line 67: Worth to create this object once and make it modifieable from the outside?

lib/frame_window_filtering/strict_labeling.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "strict_labeling.h"
+
+#include <set>
+
+namespace MouseTrack {
+
+FrameWindow StrictLabeling::operator()(const FrameWindow &window) const {
+  FrameWindow w = window;
+  // ignore labels?
+  std::set<size_t> ignore;
```
Fluci: move to be member

lib/generic/point_cloud.cpp
```diff
@@ -32,33 +33,60 @@ const PointCloud::ConstantPoint PointCloud::operator[](size_t i) const {
   return PointCloud::ConstantPoint(*this, i);
 }
 
-const Eigen::Vector3d PointCloud::min() const {
-  double minX = std::numeric_limits<double>::max();
-  double minY = std::numeric_limits<double>::max();
-  double minZ = std::numeric_limits<double>::max();
+int PointCloud::labelsDim() const {
+  // temporary approximation
+  return _labels.empty() ? 0 : _labels[0].size();
```
Fluci: a bit messy, any way to store this as a member?

lib/generic/point_cloud.h
```diff
@@ -84,8 +92,25 @@ class PointCloud {
     /// Read access to color intensity.
     ColorChannel intensity() const;
 
-    /// Convert to dx1 Eigen Matrix
-    Eigen::VectorXd eigenVec() const;
+    /// Copies labels
+    ///
+    /// 0: This point does not hold this label (non existing entries are to
```
Fluci: set to zero?

lib/generic/write_point_cloud.cpp
```diff
@@ -32,4 +32,20 @@ void write_point_cloud(const std::string &path, const PointCloud &cloud) {
   out.close();
 }
 
+void write_point_cloud_metrics(const std::string &path,
+                               const PointCloud &cloud) {
+  Eigen::VectorXd posCog;
+  posCog.setZero(3, 1);
+  for (int i = 0; i < cloud.size(); ++i) {
```
Fluci: add to point cloud as method?

lib/registration/disparity_registration_cpu_optimized.cpp
```diff
@@ -37,7 +37,8 @@ operator()(const FrameWindow &window) const {
 
   // collect data in "frame" vectors
   std::vector<Mat> framePoints(frames.size());
-  std::vector<std::vector<double>> frameIntensities(frames.size());
+  // point intdex to (x,y) coordinate
```
Fluci: typo

lib/classifier/knn.cpp
Fluci: Comparison: sci-kit uses a `predict` and `predict_prob`. Though, `predict_prob` might not be defined for an estimator (for example the svm classifier). In our case, we are only interested in "predict_prob". Fortunately, there is a very simple conversion from `predict` to `predict_prob` and back:

`predict`-> `predict_prob`: set all probabilities to zero except for the labeled predicted, which is set to zero.

`predict_prob` -> `predict`: pick label with largest probability

Using both methods in the interface seems like the most flexible solution.

lib/frame_window_filtering/background_subtraction.cpp
Fluci: No, the source code has a template without arguments setting the entire expression to zero. No change needed.

lib/frame_window_filtering/hog_labeling.cpp
Fluci: Probably not that straight forward yet? What if pyramids are used? What if a detector works by actually computing HOG's? Just leave it this way for the moment.

lib/frame_window_filtering/hog_labeling.cpp
Fluci: peak normalization is demonstrably worth it. I'll leave accross label normalization for the moment with an member option

lib/frame_window_filtering/hog_labeling.cpp
Fluci: This would definitely clean up as less variables would be floating around, but it would also expose openCV (and hence the implementation) to the interface, which might not be desirable

PR #52, opened 2018-05-29
Feature/fs/hog

Fluci:
All changes that I had to rush in during the last 1-2 weeks plus some refactoring.

### DIFF
--- a/app/pipeline_factory.cpp
+++ b/app/pipeline_factory.cpp
@@ -10,6 +10,7 @@
 #include "frame_window_filtering/disparity_gaussian_blur.h"
 #include "frame_window_filtering/disparity_median.h"
 #include "frame_window_filtering/disparity_morphology.h"
+#include "frame_window_filtering/hog_labeling.h"
 
 #include "matching/nearest_neighbour.h"
 
@@ -358,6 +359,10 @@ PipelineFactory::getWindowFiltering(const std::string &target,
     FrameWindow cageWindow = (*reader)(desiredFrame);
     auto ptr = std::make_unique<BackgroundSubtraction>();
     ptr->cage_frame(cageWindow);
+    return ptr;
+  }
+  if (target == "hog-labeling") {
+    auto ptr = std::make_unique<HogLabeling>();
     return ptr;
   }
   if (target == "none") {
--- a/app/pipeline_factory.cpp
+++ b/app/pipeline_factory.cpp
@@ -515,6 +515,7 @@ PipelineFactory::getClustering(const op::variables_map &options) const {
     return ptr;
   }
   if (target == "label-clustering") {
+    BOOST_LOG_TRIVIAL(debug) << "Creating labelClustering";
     return std::make_unique<LabelClustering>();
   }
 
--- a/app/pipeline_factory.cpp
+++ b/app/pipeline_factory.cpp
@@ -14,6 +14,7 @@
 #include "frame_window_filtering/disparity_median.h"
 #include "frame_window_filtering/disparity_morphology.h"
 #include "frame_window_filtering/hog_labeling.h"
+#include "frame_window_filtering/strict_labeling.h"
 
 #include "matching/nearest_neighbour.h"
 
@@ -364,6 +365,9 @@ PipelineFactory::getWindowFiltering(const std::string &target,
     auto ptr = std::make_unique<BackgroundSubtraction>();
     ptr->cage_frame(cageWindow);
     return ptr;
+  }
+  if (target == "strict-labeling") {
+    return std::make_unique<StrictLabeling>();
   }
   if (target == "hog-labeling") {
     if (options.count("hog-labeling-train") == 0) {
--- a/app/pipeline_factory.cpp
+++ b/app/pipeline_factory.cpp
@@ -28,6 +28,7 @@
 #include "point_cloud_filtering/subsample.h"
 
 #include "clustering/kmeans.h"
+#include "clustering/label_clustering.h"
 #include "clustering/mean_shift.h"
 #include "clustering/mean_shift_cpu_optimized.h"
 #include "clustering/single_cluster.h"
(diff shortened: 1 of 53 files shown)


app/pipeline_writer.cpp
```diff
@@ -77,14 +104,158 @@ void PipelineWriter::newFrameWindow(FrameNumber f,
...
+  for (StreamNumber s = 0; (size_t)s < window->frames().size(); ++s) {
+    maxLabelCount = std::max(maxLabelCount, window->frames()[s].labels.size());
+  }
+  auto colors = nColors(maxLabelCount);
+  // write max-label image
+  for (StreamNumber s = 0; (size_t)s < window->frames().size(); ++s) {
+    const Frame &frame = window->frames()[s];
+    int rows = frame.labels[0].rows();
+    int cols = frame.labels[0].cols();
+    // ignore background
```
Fluci, line 167: old comment

app/pipeline_factory.cpp
```diff
@@ -360,6 +366,55 @@ PipelineFactory::getWindowFiltering(const std::string &target,
...
     return ptr;
   }
+  if (target == "strict-labeling") {
+    return std::make_unique<StrictLabeling>();
+  }
+  if (target == "hog-labeling") {
+    if (options.count("hog-labeling-train") == 0) {
+      BOOST_LOG_TRIVIAL(info) << "HOG labeling needs training data, please "
+                                 "provide a path via "
+                                 "--hog-labeling-train=<path>";
```
luhug, line 376: Does this work like this, or does it need << operators in between?

luhug (approved):
Very good work Felice, thank you!
I've read through it fairly quickly and I didn't spot any errors.. I don't 100% understand everything, I would like to of course, but I have more important things to spend my time on right now.. As long as it compiles and the tests pass I'm happy.

examples/aggressive_gauss.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
itko, line 1: Interesting approach to use these scripts. Is this for Tilo  or for us?

If it's for Tilo, it's not obvious exactly what each script is for. Aggressive Gauss vs. Baseline Benchmark, etc.

That being said, I'm not sure these examples will serve much purpose for Tilo  anyway, so perhaps a simple statement in the README that mentions their existence is good enough.

lib/frame_window_filtering/hog_labeling.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "frame_window_filtering.h"
+
+#include "classifier/classifier.h"
+
+#include <memory>
+
+namespace MouseTrack {
+
+class HogLabeling : public FrameWindowFiltering {
```
itko: Would you really consider this a filtering operation?

lib/frame_window_filtering/hog_labeling.h
```diff
@@ -0,0 +1,54 @@
...
+  void train(const Mat &X_train, const Vec &y_train);
+  virtual FrameWindow operator()(const FrameWindow &window) const;
+
+private:
+  std::unique_ptr<Classifier> _classifier;
+  // highest label + 1: we need to know, how many labels there are
+  int _numLabels;
+
+  /// width of the sliding window
+  int _windowWidth = 64;
```
itko: Shouldn't we be able to set all these parameters? For example, 64 is quite an arbitrary value we chose

itko (changes requested):
I think we need quite a bit more comments if we hope for someone to be able to improve this in the future

That being said, this is lower priority right now, so I think you should merge anyway

app/pipeline_factory.cpp
Fluci: Weird C syntax, this actually works. (I don't like the style either, but clang-format "enforces" it.)

examples/aggressive_gauss.sh
Fluci: Agreed, I'll add comments in the readme and in each script to make clear what they do.

lib/frame_window_filtering/hog_labeling.h
Fluci: Yes. "Filtering" is probably a bad term, I think of it as an operation that takes a frameWindow and returns a modified copy. As always, open for a better terminology.

lib/frame_window_filtering/hog_labeling.h
Fluci: Agreed, but I'm not sure what/how to expose them.
I'm a bit afraid of just exposing everything, as I have no idea for constraints between parameters (example: does window width need to be a multiple of cell width?).

I think, for a first iteration, I'll add the "necessary" parameters like "windowWidth/Heigh" and "stepSize" such that we can play around with them.

itko (approved):

lib/frame_window_filtering/hog_labeling.cpp
```diff
@@ -30,6 +30,34 @@ std::vector<cv::Point> slidingWindows(int imgWidth, int imgHeight, int stepSize,
   return result;
 } // namespace MouseTrack
 
+
+int HogLabeling::slidingWindowWidth() const {
```
itko: Thanks for updating these, but what about block size/stride? I would suggest either parameterizing them or at least defining them based on the window parameters.

Also, setters/getters are good, but why not just initialize everything in the constructor? Would these parameters be modified dynamically?

itko (changes requested):
After addressing the comment (if you agree with it) please just merge. No need to wait.

lib/frame_window_filtering/hog_labeling.cpp
Fluci: As I understand it, the patch from which we extract the hog feature vector always has the same size as the sliding window? Or do you mean the cells of a patch? Until now, patchSize=slidingWindowSize and stride were the only values we've been tweaking, if it shows useful to change other things, I'll add accessors later.

Setter/Getter vs. Constructor: 

Probably mostly a design choice, I've never been taught anything official about this. The constructor has the responsibility to create a valid object. This implies, that all information needed to create a valid object, has to be passed to the constructor. 
One design rule I try to follow, is to always put "reasonable" default values if available. Hence when there's a good default value, I don't see a point in putting it in the constructor and instead make it available as property one can change.

As always, I think there are two sides: It is probably slightly more efficient to pass stuff to the constructor as you only have to write the value once. On the other hand side I think it is more important to have a flexible and robust design. Experience taught me, that complicated constructors can be very annoying, for example when you write a factory for the class.
As this module's constructor is, by design, called "rarely", I prefer robust design over efficiency.

PR #53, opened 2018-06-04
Feature/fs/spatial refac

Fluci:
As announced, I've refactored the SpatialOracle interface: Instead of taking a single point, it now takes a batch of query points. This allows us to use FLANN's parallelization feature (implemented) and further optimizations for other objects are possible (Sorting points into cells and then only querying the cells could give a fairly descent speed up for the UniformGrid, but that's not implemented yet). 

Since it's an interface change, it touches quite some files.


itko (approved):
Quickly looked over. Approved.

luhug (approved):

PR #55, opened 2018-06-12
Background subtraction dynamic thresholding

luhug:
Changed the background-subtraction-threshold parameter such that it works like in itto's python notebook. This also includes a newer version of the validation script because I needed it for verification.


Fluci (approved):

lib/frame_window_filtering/background_subtraction.h
```diff
@@ -25,7 +25,7 @@ class BackgroundSubtraction : public FrameWindowFiltering {
 
 private:
   FrameWindow _cage_frame = FrameWindow();
-  double _threshold = 0.01;
+  double _threshold = 1;
```
itko: Would change name to something more descriptive. It's not a threshold it's a factor multiplied by the threshold. "_factor" or something makes more sense.

itko (changes requested):

